PR #149, opened 2023-03-08
Add `repository` key to keystatic/core package.json

jasikpark:
Similar to https://github.com/withastro/astro/blob/main/packages/astro/package.json, this allows tools like https://docs.renovatebot.com/ to find the source for the package

image

^ right now this just diffs the zip output hosted on npm

This could also be done for the other packages in the monorepo, I just wanted to push an MVP while I'm being lazy and editing in the Github UI 😅


jasikpark:
no changeset needed, this is a documentation change

JedWatson:
Thanks @jasikpark

I added a changeset as well, since I think without actually releasing the packages, renovate won't know where to go to find the new package.json files with the repository in them ... bit of a catch-22

PR #247, opened 2023-04-19
Renaming Voussoir to Keystar UI

JedWatson:
Largely renames Voussoir to Keystar UI, which better explains what these components are doing here 🙂 

(for anyone curious, the DS will eventually be used by both Keystatic and Keystone, so we're playing on the `key*` theme)

I've tested this locally, and created the new npm scope, so I think it's good to merge but we'll need to make sure the release bot has access and also update the docs deployment on Vercel to the new project name.

Oh, and the Changeset bot didn't detect the new packages correctly so I'll need some help getting the release set up correctly in the morning 🤞🏻


emmatown:
This PR is just gonna get stale, we'll do this in a different PR

PR #263, opened 2023-05-10
Increase gap between fields in data entry form

simonswiss:
Increase the gap value to `xlarge` between fields in a data entry form.


emmatown (changes requested):
This also needs to be changed for conditional fields

emmatown (approved):

PR #270, opened 2023-05-15
Docs - Layout and responsive navigation

jtbrolin:
- Added mobile navigation (hamburger button + slideout)
- Splitted out navigation.tsx into smaller files in components/navigation for easier consumption
- Separate RootLayouts for the frontend and the Keystatic Admin UI using route groups, `(frontend)` + `(keystatic)`


docs/src/hooks/useLockedBody.ts
jtbrolin, line 1: Grabbed from https://usehooks-ts.com/react-hook/use-locked-body

docs/src/components/navigation/social-links.tsx
```diff
@@ -0,0 +1,83 @@
...
+    'shrink-0 rounded-lg p-3 hover:bg-keystatic-gray-light active:bg-keystatic-gray transition-colors';
+
+  return (
+    <div className="flex gap-2">
+      <Link
+        href="https://fosstodon.org/@keystatic"
+        className={styles}
+        tabIndex={tabIndex}
+      >
+        <span className="sr-only">Keystatic on Mastodon</span>
```
BarnabyBishop, line 14: `sr-only` - neat!

docs/src/components/navigation/toc.tsx
```diff
@@ -0,0 +1,38 @@
+export function TableOfContents() {
+  return (
+    <div className="w-[12rem] sticky top-10 lg:top-32 self-start hidden md:block lg:hidden 2lg:block">
+      <h5 className="text-xs uppercase text-black text-stone-500">
```
BarnabyBishop, line 4: Got some css double up with `text-` here

docs/src/content/navigation.yaml
```diff
@@ -1,7 +1,11 @@
 navGroups:
   - groupName: Getting started
     items:
-      - label: Automated setup
+      - label: Introduction
+        link:
+          discriminant: page
+          value: introduction
+      - label: Automated Setup
```
BarnabyBishop, line 8: Should we use sentence or title here? Seems to be a bit of a mix (topical topic for today)

BarnabyBishop (changes requested):
Looking shmick. I need to look at `useLockedBody` more carefully but that can wait until Monday

docs/src/components/navigation/mobile-nav.tsx
```diff
@@ -0,0 +1,139 @@
...
+        onScroll={onScrollHandler}
+        className={`overflow-y-auto list-none fixed lg:hidden top-0 bg-white h-screen w-64 z-30 drop-shadow-2xl flex flex-col transition-[right] duration-300 ${
+          navOpen ? '-right-0' : '-right-full'
+        }`}
+      >
+        <FocusLock disabled={!navOpen}>
+          {/* Sticky close */}
+          <div
+            className={`sticky top-0 left-0 right-0 p-2 bg-white z-30 justify-end flex transition-shadow ${
+              scrollTop === 0 ? 'shadow-none' : 'shadow-lg'
```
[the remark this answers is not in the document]
jtbrolin: Ahh amazing testing and suggestion!! Thanks 🙏🏻

docs/src/components/navigation/mobile-nav.tsx
```diff
@@ -0,0 +1,139 @@
...
+          navOpen
+            ? 'visible opacity-100'
+            : 'invisible opacity-0 lg:visible lg:opacity-100'
+        }`}
+      />
+
+      {/** Slideout menu */}
+      <ul
+        onScroll={onScrollHandler}
+        className={`overflow-y-auto list-none fixed lg:hidden top-0 bg-white h-screen w-64 z-30 drop-shadow-2xl flex flex-col transition-[right] duration-300 ${
```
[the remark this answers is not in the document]
jtbrolin: Ahh great spotting I'll look into that!

docs/src/components/navigation/social-links.tsx
jtbrolin: Can't take the credit for this one unfortunately :D

docs/src/components/navigation/toc.tsx
jtbrolin: Ahhh good catch 🙌🏻

docs/src/content/navigation.yaml
jtbrolin: Hmm yes.. I'll prob give Tuan a shout re: this and see

jtbrolin: Fixed, FYI we're following https://balance.reckon.com/content/capitalisation

jtbrolin:
- [ ] Test that `dvh` unit doesn't break the layout in unsupported browsers

jtbrolin:
> Looking shmick. I need to look at `useLockedBody` more carefully but that can wait until Monday

@BarnabyBishop Yeah please have a look on that one, I pretty much directly took it from [usehooks](https://usehooks-ts.com/react-hook/use-locked-body), mostly because it had the scrollbar logic included. But haven't tested this (on Windows for example to see how bad it is if the scrollbars disappears on "lock" and we don't have this logic)

docs/src/components/navigation/keystatic-logo.tsx
```diff
@@ -0,0 +1,27 @@
+import Link from 'next/link';
+
+export function KeystaticLogo() {
+  return (
+    <Link href="/">
```
ChrisLaneAU, line 5: We should probably put an `aria-label` on here. What do you think about: `Home of Keystatic`?

ChrisLaneAU: And maybe move the link out to the parent?

docs/src/components/navigation/header-nav.tsx
```diff
@@ -0,0 +1,44 @@
...
+      label: string;
+      href: string;
+      title: string | undefined;
+    }[];
+  }[];
+};
+
+export function HeaderNav({ navigationMap }: NavProps) {
+  const pathname = usePathname();
+  const isDocsNav = pathname?.includes('/docs');
```
ChrisLaneAU, line 22: Maybe a bit pedantic but what do you think about `pathname?.startsWith('/docs')? Just so we don't get something like `/templates/docs-template` etc?

docs/src/hooks/useLockedBody.ts
```diff
@@ -0,0 +1,54 @@
...
+      }
+    };
+  }, [locked]);
+
+  // Update state if initialValue changes
+  useEffect(() => {
+    if (locked !== initialLocked) {
+      setLocked(initialLocked);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
BarnabyBishop, line 48: Spent some time looking at this hook and I'm not a fan of this `useEffect`. Not listing the deps, especially `locked` and `setLocked` could cause some possible bugs. 

I'd suggest removing this whole useEffect as it is listening for changes to `initialValue` and updates the state, but if it is doing that it is no longer 'initialValue' any more, it is just regular state! 

We aren't changing this value dynamically anyway, so safe to remove without causing any issues for our current use-case.

docs/src/components/navigation/header-nav.tsx
BarnabyBishop: Good thinking

docs/src/app/privacy-policy/page.tsx
```diff
@@ -1,15 +1,13 @@
-import Navigation from '../../components/navigation';
 import PrivacyPolicy from '../../components/privacy-policy';
 import Footer from '../../components/footer';
 
 export default function PrivacyPolicyPage() {
   return (
-    <div className="min-h-screen">
-      <Navigation />
+    <>
       <main>
```
ChrisLaneAU, line 7: Do we want to move the `<main>` up to `/docs/layout.tsx` so that we don't have to keep track of what has a `main` element?

docs/src/app/privacy-policy/page.tsx
```diff
@@ -1,15 +1,13 @@
...
         <PrivacyPolicy />
       </main>
       <Footer />
```
ChrisLaneAU, line 10: Similar question with the `<Footer />`?

docs/src/components/footer.tsx
```diff
@@ -1,23 +1,50 @@
...
+      </div>
+
+      <div className="mx-auto mt-2 flex max-w-7xl flex-col justify-between gap-y-4 gap-x-12 px-6 text-center md:flex-row">
+        <p className="leading-tight text-keystatic-gray-dark text-center md:text-left">
+          <CopyrightPrivacyPolicyContent />
         </p>
-        <p className="leading-none text-keystatic-gray-dark">
+
+        <p className="leading-tight text-keystatic-gray-dark text-center md:text-right">
           Made with <span className="text-[#ff3838]">♥</span> in Australia by a
```
ChrisLaneAU, line 21: Running the screen reader over this I think we should change it to ❤ so that it doesn't read out the playing card thing.

docs/src/app/docs/page.tsx
```diff
@@ -3,23 +3,7 @@ import Button from '../../components/button';
...
-          >
-            <path
-              fill="#F7DE5B"
-              d="m4.239.201 92.684 2.883 100.722 7.097 99.043 7.211 94.363 2.77-21.813 9.088 14.042 9.919 2.873 8.7-14.795 6.043 7.844 5.477 7.843 5.476-14.691 6.037 11.104 9.535 3.927 10.77-93.59-1.7-100.082-5.647-116.75-3.055-76.39-9.559 12.857-8.312-11.94-9.45 5.534-10.258-4.618-7.502 16.812-1.055L7.21 20.478l5.332-11.703L4.239.201Z"
-            />
-          </svg>
-          <span className="relative">Docs</span>
-        </span>
-      </h2>
+      <h1 className="text-3xl font-bold sm:text-4xl">Keystatic Docs</h1>
```
ChrisLaneAU, line 6: What's the plan for this page? More to come? I can see line 10 is using markdown.

BarnabyBishop: There's a ticket for that, I should get round to it eventually 🥴

https://linear.app/keystatic/issue/KEY-119/docs-homepage

I believe the current plan is to redirect this page to a content page called `Installation` which helps users on the first step of their keystatic journey.

docs/src/components/footer.tsx
BarnabyBishop: Great catch

docs/src/app/docs/page.tsx
ChrisLaneAU: Awesome!

docs/src/components/navigation/mobile-nav.tsx
```diff
@@ -0,0 +1,139 @@
...
+  }, [setLocked]);
+
+  const openNav = useCallback(() => {
+    setNavOpen(true);
+    setLocked(true);
+  }, [setLocked]);
+
+  // Close the mobile menu when pathname (route) changes
+  // https://nextjs.org/docs/app/api-reference/functions/use-pathname#do-something-in-response-to-a-route-change
+  useEffect(() => {
```
ChrisLaneAU, line 30: Should we add a listener for `esc` and close on press?

docs/src/components/navigation/mobile-nav.tsx
```diff
@@ -0,0 +1,139 @@
...
+    event: UIEvent<HTMLUListElement> & { target: HTMLUListElement }
+  ) => {
+    const { target } = event;
+    setScrollTop(target.scrollTop);
+  };
+
+  return (
+    <>
+      {/** Hamburger menu button */}
+      <Button
```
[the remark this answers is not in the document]
ChrisLaneAU: Actually, it reads pretty nicely. The only thing I was really curious about was whether we should wrap this button in a `nav` that includes the menu? Or a `role="navigation"` wrapper? 

The other thing that we should probably do is focus the `Menu` button when the menu gets closed.

docs/src/components/navigation/social-links.tsx
```diff
@@ -0,0 +1,83 @@
...
+              clipRule="evenodd"
+            />
+          </g>
+          <defs>
+            <clipPath id="a">
```
ChrisLaneAU, line 30: Is that `id` correct? Line 75 has the same id. I'm thinking these need a unique id from `useId`?

docs/src/hooks/useIsomorphicLayoutEffect.ts
ChrisLaneAU, line 1: I don't think we need this hook anymore. [React has stopped printing this warning](https://github.com/facebook/react/pull/26395).

docs/src/utils/reader.ts
ChrisLaneAU, line 1: 😍

ChrisLaneAU (changes requested):
Looking good, just some small changes in isolated comments. Sing out if any of it doesn't make sense or you believe is inaccurate.

docs/src/hooks/useIsomorphicLayoutEffect.ts
BarnabyBishop: Has this been released yet though? Looks like the latest react version was published almost a year ago: https://github.com/facebook/react/releases

ChrisLaneAU: Oh good point. I just assumed. I wonder if it's actually just because you can't use hooks on the server, which we're not. Our use is only within the `header-nav.tsx`, which is a client component.

So, my newfound understanding is that we don't need this hook because you can't use hooks on the server (Next will throw) and the only reason it existed was to circumvent the warning. Does that sound plausible?

docs/src/components/navigation/header-nav.tsx
jtbrolin: Great catch – thank you!

docs/src/hooks/useLockedBody.ts
jtbrolin: Ahh great catch! And thanks for the explanation <3

docs/src/app/privacy-policy/page.tsx
jtbrolin: Yeah it's not optimal.. I went down this route because the landing pages `/` + `/thank-you` etc and `/docs` have different layouts and different looking <footer>s, and this way the <nav> and <footer> on /docs-pages are not inside the <main> element. So I struggled to figure out a better way..

But maybe nesting <nav> inside <main> on /docs is fine, since it's a "sub menu" anyways (same goes for the TOC which I was wondering if it should be an <aside>)..

So then I'd only have to solve for the different footers 🤔 which the only solution I can come up with is: 
In the <Footer /> component, do the same `pathname?.startsWith('/docs');` to not render any markup from the `/docs/src/app/layout.tsx` if it's `true`.. Hmmm what you reckon?

docs/src/components/footer.tsx
jtbrolin: Wow didn't know it did that 😅 Thank you!!

docs/src/components/navigation/mobile-nav.tsx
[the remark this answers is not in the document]
jtbrolin: > this button in a nav that includes the menu? Or a role="navigation" wrapper?
Yeah, it's not very clear but this all gets rendered inside a `<nav>` already (from the header-nav.tsx), and adding role="navigation" just doubles up a bit when testing with VoiceOver but I'm unsure of what approach is "right" you know.. 

> focus the Menu button when the menu gets closed.
Can defo do that!

docs/src/components/navigation/social-links.tsx
jtbrolin: Ahh good catch, this is all copypaste from old, so haven't even noticed that, but yes will sort 🙌🏻

docs/src/hooks/useIsomorphicLayoutEffect.ts
jtbrolin: Awesome – good to know, thanks for that intel/research!! Will change to useLayoutEffect

jtbrolin:
- [ ] pathname?.startsWith('/docs')
- [ ] /keystatic-logo-link and aria-label / aria-hidden
- [ ] remove useEffect from useLockedBody
- [ ] main and footer in layout.tsx ?
- [ ] change it to red heart emoji
- [ ] focus the Menu button when the menu gets closed.
- [ ] add a listener for esc and close on press?
- [ ] social links unique path id’s
- [ ] useLayoutEffect rather than useIsomorphicLayoutEffect.
- [ ] switch 2lg to rems

BarnabyBishop (approved):
Noice

docs/src/app/(keystatic)/layout.tsx
BarnabyBishop, line 1: You can remove this `layout.tsx` as keystatic already has one in `/app/(keystatic)/keystatic/layout.tsx`, it doesn't need one down here. 

Deleting this actually fixes some of the weird textbox colours we are seeing in the Admin UI which I now know was introduced when I created the original layout.tsx 😬

BarnabyBishop (changes requested):
One tiny change 🤏

docs/src/app/(keystatic)/layout.tsx
jtbrolin: Ahh ok will try :)

ChrisLaneAU (approved):
Love it! Well thought out, great job :)

BarnabyBishop (approved):
Double approved ✅  ✅

ChrisLaneAU (approved):
Thank you!

PR #271, opened 2023-05-15
Keystatic docs site

BarnabyBishop:
Still a bit of a WIP, but this PR contains the new Keystatic website, with docs which are managed in its own keystatic setup.

- Copied keystatic-site into the docs folder
- Added `/docs` section to the docs site with content editable in keystatic
- Added a bunch of content
- Changed voussoir npm scripts from `docs` to `designsystem`, e.g. `npm run build:docs` -> `npm run build:designsystem` so we could steal the docs commands (sorry Joss). I'll update the voussior vercel app with this change before merging.


docs/public/images/templates/blog@2x.png
jtbrolin, line 1: Just noticed that the @1x and @2x versions are different from the @3x version, and looks like @3x is the only one being used

docs/README.md
```diff
@@ -0,0 +1,14 @@
+# keystatic-deploy
+
+To submit forms to airtable, you need to add a `.env` file in the project root with the following environment variables:
```
ChrisLaneAU, line 3: Should we have a `.env.example`?

docs/next.config.js
```diff
@@ -0,0 +1,9 @@
+/** @type {import('next').NextConfig} */
+module.exports = {
+  reactStrictMode: false,
+  typescript: { ignoreBuildErrors: true },
+  experimental: {
+    appDir: true,
```
ChrisLaneAU, line 6: Do we still need this now that the app router is stable? I don't actually know 🤷

docs/public/images/templates/blog@2x.png
ChrisLaneAU: I can fix this in my images PR + ancillary tasks

docs/public/safari-pinned-tab.svg
ChrisLaneAU, line 1: Is this image correct?

emmatown (changes requested):
I'll probably have thoughts on a bunch of this but I can review it async. I just have one blocking concern: the fonts and images bloating this repo. For the site icons, just have a single `icon.svg` in the `app` directory. For the other images, could you put them in an R2 bucket or something? For the font, use `next/font`.

docs/src/app/docs/page.tsx
```diff
@@ -0,0 +1,91 @@
+import Button from '../../components/button';
+
+export default function Docs() {
+  return (
+    <>
+      <h2 className="text-3xl font-bold sm:text-4xl">
```
ChrisLaneAU, line 6: Should this be `h1`?

emmatown:
For now, have a custom block with a text field for the image url

BarnabyBishop:
Closing the PR as we no-longer need it right now. 

All PR feedback has been put in Linear.

PR #273, opened 2023-05-17
Docs - move existing pages into the app directory

BarnabyBishop:
Docs: Moved all pages from pages to docs

- Changed useRouter from next/router to next/navigation
- Updated modal logic to use useRouters simpler API
- Removed next-seo as Next13 provides that out of the box
- Added a .env.example file

Tested:
- Creating a new project from the template
- Filling in the form when creating a new project from the template
- 'Get notified' when a template becomes available
- Adding email for updates
- Sending a message


docs/src/app/docs/layout.tsx
```diff
@@ -37,66 +37,57 @@ export default async function RootLayout({
   }));
 
   return (
-    <html lang="en">
-      <body>
```
BarnabyBishop, line 41: Despite the big diff the only change here _should_ be the removal of the html and body tags since we have a new layout higher up in the app directory.

jtbrolin (commented):
Looking awesome 😍 but I noticed that when opening a modal the page scrolls to the top, are you getting that behaviour as well @BarnabyBishop?

https://github.com/Thinkmill/keystatic/assets/82560714/67fcf381-2a71-4bb2-9ce2-1205a0e7daaa

docs/src/components/mailing-list.tsx
```diff
@@ -1,3 +1,5 @@
+'use client';
```
ChrisLaneAU, line 1: Since this component doesn't use any hooks, should we move this directive down to the `MailingListForm`?

emmatown: I'd recommend leaving the `use client` here so that e.g. if we have this mailing list component on many pages, when users do a client-side navigation, all the markup in this component doesn't have to be downloaded again

ChrisLaneAU: Ahh nice, that makes sense, thanks!

docs/src/components/templates.tsx
```diff
@@ -127,14 +117,18 @@ function useLastTruthyValue<T>(val: T): T {
...
 
   const lastTemplate = useLastTruthyValue(templateMatch);
 
   const onClose = () => {
-    const { template: _template, ...restQuery } = router.query;
-    router.push({ query: restQuery }, undefined, { scroll: false });
+    // Remove the template search param
+    const queryParams = new URLSearchParams(searchParams || '');
+    queryParams.delete('template');
+    router.push(`/?${queryParams.toString()}`);
```
ChrisLaneAU, line 131: Should be able to remove the `.toString()` here since we're already stringifying with interpolation.

docs/src/components/templates.tsx
```diff
@@ -178,11 +172,9 @@ export default function Templates() {
...
                     template={template}
                     onClick={() => {
-                      router.push(
-                        { query: { ...router.query, template: template.id } },
-                        undefined,
-                        { scroll: false }
-                      );
+                      const params = new URLSearchParams(searchParams || '');
+                      params.append('template', template.id);
+                      router.push(`/?${params.toString()}`);
```
ChrisLaneAU, line 177: Same here with removing `.toString()`.

docs/src/components/templates.tsx
BarnabyBishop: Nice pickup

ChrisLaneAU:
> Looking awesome 😍 but I noticed that when opening a modal the page scrolls to the top, are you getting that behaviour as well @BarnabyBishop?

I'm seeing this, too. Maybe we need to do `router.back()` on close to restore the scroll position (not amazing UX) or pass this state a different way?

ChrisLaneAU:
> One of the reasons to use a query string for modal state is so that the modal is shareable as a link
Yep, that makes sense. Also on mobile it takes up such a large amount of the screen, it could be seen as a separate page.

>  I tested it for fun anyway and it doesn't preserve the scroll position anyway, which is also a known issue with the router
That's funny, it did work for me 🤷  but I think you're right `back` is not a good solution.

> Perhaps it is worth it as the scroll issue seems fairly noticeable.
What about focusing the CTA after the push back to `/`? I can't see that it's possible to pass state with `next/navigation` outside query params so I'm not sure if this would work. Maybe we don't need to pass state, though and can keep it locally? So maybe this would work?:
```
router.push(`/?${queryParams}`);
ctaRefs.current[template.id].focus();
```

Also, there is this [custom router events solution](https://nextjs.org/docs/app/api-reference/functions/use-router#router-events) but I doubt it's worthwhile just for this.

I noticed that `<Link scroll={false} />` didn't work either 🤷

jtbrolin (approved):
Nice one! 🙌🏻

BarnabyBishop:
> What about focusing the CTA after the push back to `/`?

I had a play with this idea, the issue is when the modal opens you still jump to the top of the page, then you jump down when closing. 

Instead I did a _bit_ of a workaround:
- The modal state is now controlled by internal component state, no more scrolling when opening/closing
- Now using `history.pushState` to update the URL so it is sharable
- On entry to the page with the query param it will show the modal

The caveat is that since we are going outside of Next's router, the back/forward buttons don't always cause the modal to show. Not ideal but seems like the least-bad solution until Next fixes the scroll issue in their router, then we can go back to URL state.

ChrisLaneAU:
> I had a play with this idea, the issue is when the modal opens you still jump to the top of the page, then you jump down when closing.
> 
> Instead I did a _bit_ of a workaround:
> 
> * The modal state is now controlled by internal component state, no more scrolling when opening/closing
> * Now using `history.pushState` to update the URL so it is sharable
> * On entry to the page with the query param it will show the modal
> 
> The caveat is that since we are going outside of Next's router, the back/forward buttons don't always cause the modal to show. Not ideal but seems like the least-bad solution until Next fixes the scroll issue in their router, then we can go back to URL state.

Yep, agreed. Lovely work 🙌

docs/src/components/templates.tsx
```diff
@@ -126,15 +122,25 @@ function useLastTruthyValue<T>(val: T): T {
...
   const onClose = () => {
-    const { template: _template, ...restQuery } = router.query;
-    router.push({ query: restQuery }, undefined, { scroll: false });
+    // Remove the template search param
+    const queryParams = new URLSearchParams(searchParams || '');
+    queryParams.delete('template');
+    history.pushState(
+      {},
+      '',
+      Array.from(queryParams.entries()).length ? `?${queryParams}` : '/'
```
ChrisLaneAU, line 140: Could do `queryParams.size ? ...` but since we're throwing this away at some point, probably not worth it?

ChrisLaneAU (approved):
Nice work!

docs/src/components/templates.tsx
BarnabyBishop: Yessss so much better. `size` didn't come up in intellisense. That's what I get being too lazy to open the docs 🫣

BarnabyBishop: Ah, that's why it doesn't come up in intellisense:
image

https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/size#browser_compatibility

Thanks to our TS CI step for alerting me to that 😅

ChrisLaneAU: Ahh interesting - TIL. iOS Safari yikes. I've defs used this places I shouldn't 😅

ChrisLaneAU: So do you know if there's a VS extension/config I can apply the same CI rule?

PR #274, opened 2023-05-19
Image storage in Keystatic Cloud and custom docs image component

ChrisLaneAU:
This PR:
1. Adds a custom image component called `Cloud image` to add cloud hosted images
2. Updates the `installation-astro` guide to use Keystatic cloud images


docs/src/components/previews/CloudImagePreview.tsx
```diff
@@ -0,0 +1,30 @@
+'use client';
+
+import { useState } from 'react';
+import { ProgressCircle } from '@voussoir/progress';
+import { getDefaultSrcSet } from '../../utils';
+import { CONTENT_MAX_WIDTH_DESKTOP } from '../../constants';
+
+export function CloudImagePreview(props: any) {
```
BarnabyBishop, line 8: I wonder how hard it would be to type this? Using something like `InferRenderersForComponentBlocks` but with preview props?

docs/src/components/document-renderer.tsx
```diff
@@ -111,4 +113,18 @@ const componentBlockRenderers: InferRenderersForComponentBlocks<
       </div>
     );
   },
+  'cloud-image': ({ href: src, alt, sizes, height, width }) => {
```
BarnabyBishop, line 116: The Keystatic field has `srcSet` as an option. Should we use that here if provided, or remove it from the UI config?

docs/keystatic.config.tsx
```diff
@@ -45,6 +46,36 @@ export const componentBlocks = {
...
+          },
+        },
+      }),
+      alt: fields.text({
+        label: 'Alt text',
+      }),
+      height: fields.text({
+        label: 'Height',
+      }),
+      width: fields.text({
```
jtbrolin, line 67: Could potentially be a `fields.integer` cause when playing I tried putting in "12rem" which didn't work (and then I read the code 😅)

Or otherwise maybe we could set the width/height with css instead 🎨

docs/src/components/document-renderer.tsx
```diff
@@ -111,4 +113,18 @@ const componentBlockRenderers: InferRenderersForComponentBlocks<
...
+    const imgMaxWidthPx = `${parseInt(CONTENT_MAX_WIDTH_DESKTOP) * 16}`;
+
+    return (
+      <img
+        alt={alt || ''}
+        src={src}
+        height={height}
+        width={width || imgMaxWidthPx}
+        srcSet={getDefaultSrcSet({ src })}
```
jtbrolin, line 125: Should this guy have a `srcSet || getDefaultSrcSet({ src })` potentially? Or maybe the plan is to not expose this field in the future really?

docs/src/components/previews/CloudImagePreview.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+  const [isLoading, setIsLoading] = useState(!!src);
+
+  return (
+    <>
+      <p>Href: {src}</p>
+      <p>Alt text: {alt}</p>
+
+      {isLoading && <ProgressCircle aria-label="Loading…" isIndeterminate />}
+      <img
```
jtbrolin, line 21: I was playing around a little, and wondered if the preview also could have the `height` and `width` so you can see it change after you've edited and click "Done"?

docs/src/constants/index.ts
```diff
@@ -0,0 +1,17 @@
...
+ * - Content + Side Nav = `MAIN_EL_MAX_WIDTH` (tailwind `max-w-7xl`)
+ * - padding left + right = 3rem (`px-6`)
+ * - Side Nav = 15rem (`w-60`)
+ * - ON THIS PAGE = 12rem
+ * - gap = 1.5rem (`gap-6`)
+ * - padding left = 3rem (`pl-12`)
+ * -----------------------------------
+ * Total: 42rem
+ */
+export const CONTENT_MAX_WIDTH_DESKTOP = '42rem';
```
jtbrolin, line 17: Niice new shiny constants 🙌🏻

jtbrolin (commented):
Niice - just jotted down a couple thoughts I had when testing :)

docs/src/utils/index.ts
```diff
@@ -1,3 +1,13 @@
+import { DEFAULT_IMG_WIDTHS } from '../constants';
+
 export function cx(...classes: string[]) {
   return classes.filter(Boolean).join(' ');
 }
+
+export function getDefaultSrcSet({ src }: { src: string }) {
+  // When there's auth, get a list of the configured sizes for this org
+  // If they accept `*`, then fall back to this default
```
BarnabyBishop, line 9: This comment confused me a tad. 
Does `When there's auth` refer to the future, or what this function is currently doing?
I assume the former as I don't see any auth stuff. 

Perhaps "When auth becomes available from Keystatic cloud..."?

Or my personal fav: 
```
// TODO: When auth becomes available...
```
😂

BarnabyBishop (changes requested):
Lookin' good, few comments. Excited to get this one out!

docs/keystatic.config.tsx
ChrisLaneAU: Yeah good thinking, I like it 🙌 

The height and width are more to avoid layout shift by hinting to the browser what to expect and reserve that space. I was even thinking we could put these values in an `Advanced features` Accordion with some descriptions or something but I didn't want to get too far ahead of myself because this all really needs some dev/design sync + thoughts.

I will defs do the `fields.integer`, though 🙂

docs/src/components/document-renderer.tsx
ChrisLaneAU: Yes, it should!

docs/src/components/previews/CloudImagePreview.tsx
ChrisLaneAU: Yeah, good thinking. I would actually like to programatically add defs a height (if one isn't supplied) but probably also a width as well. `next/image` requires these and they will provide a much better UX. Somehow storing an aspect ratio would probably be sufficient to figure it out.

I can add the height/width indicators, though.

docs/src/components/previews/CloudImagePreview.tsx
ChrisLaneAU: Ahh yes, I will look into this.

docs/src/components/document-renderer.tsx
ChrisLaneAU: Yes, we should. I'm not sure if you want to sync on feature requirements now or later for this but I just threw a few things in I thought would be useful. I also haven't used `next/image` for reasons. If you're happy to just get something going now, maybe we can sync later?

docs/src/utils/index.ts
ChrisLaneAU: Haha you know, the last thing I did before I committed was remove the `TODO` 😂 

And yeah, you're absolutely right this refers to the future. How we integrate the config options (including the widths as picked up in this function) from the user's/project's/org's cloud storage will probably change or at least extend this. I actually now don't think we need the comment, just a sync on how cloud integration will work when we're up to that. That sound ok to you?

BarnabyBishop: Yup totally, there is a lot of unknown in how cloud images will integrate at this stage, so we can just come back to it then.

docs/src/components/document-renderer.tsx
BarnabyBishop: Let's keep it simple for now as there is so much design and tech work we need to do before we know what this will actually look like. This will do nicely for now.

docs/src/components/previews/CloudImagePreview.tsx
ChrisLaneAU: I wasn't able to do this because of the circular reference, so have just left it as `any`. I think this will be replaced fairly soon so I'm thinking it's not worth typing manually but I'm happy to if you'd prefer.

BarnabyBishop: Nah all good. Let's leave it

jtbrolin (approved):
Nice!!

BarnabyBishop (approved):
🖼️

PR #275, opened 2023-05-22
Docs - On this page

BarnabyBishop:
- Generates a table of contents from the docs page content
- Updated the table of contents component to use generated data
- Added an optional `anchor` prop to the heading component to allow it to be linkable
- Removed all headings from the docs content document editor except H2 and H3, wanted to avoid an accidental H1 double up, and I'm not sure we need H4 -> 6

@jtbrolin - this will have some minor conflicts with your branch which I'll resolve after you have merged your PR.


docs/src/components/heading.tsx
```diff
@@ -1,12 +1,19 @@
...
   children: ReactNode;
+  anchor?: boolean;
 };
 type ValidHeading = 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
 
-export default function Heading({ level, children }: HeadingProps) {
+export default function Heading({
+  level,
+  children,
+  anchor = false,
```
ChrisLaneAU, line 15: I don't think you need the `= false`, do you?

docs/src/components/navigation/table-of-contents.tsx
```diff
@@ -0,0 +1,33 @@
...
+};
+
+type TableOfContentsProps = {
+  headings: Heading[];
+};
+
+export function TableOfContents({ headings }: TableOfContentsProps) {
+  return (
+    <div className="w-[12rem] sticky top-10 lg:top-32 self-start hidden md:block">
+      <h5 className="text-xs uppercase text-stone-500">On this page</h5>
```
ChrisLaneAU, line 15: Should this be `h2`? Or `h3`, similar to the other nav?

docs/src/components/navigation/table-of-contents.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+      <ul className="mt-2 flex flex-col gap-3">
+        {headings.map(({ level, text }) => (
+          <li>
+            <a
```
ChrisLaneAU, line 20: Should we use `next/link` here?

docs/src/app/docs/layout.tsx
```diff
@@ -60,15 +60,7 @@ export default async function RootLayout({
...
-            <div className="grid gap-6 grid-cols-[auto,12rem]">
-              <div>{children}</div>
-
-              <div className="sticky top-28 self-start">
-                <h5 className="text-xs uppercase text-stone-600">
-                  On this page
-                </h5>
-              </div>
-            </div>
+            <div className="grid gap-6 grid-cols-[auto,12rem]">{children}</div>
```
ChrisLaneAU, line 63: Should we move this grid to the `[slug]/page.tsx` where the `<TableOfContents />` is used? Just thinking about that `12rem` only being used for the ToC so maybe they should stay together rather than being used e.g. on the `/docs` page?

ChrisLaneAU (commented):
Looks great! Feels like so much less work than other docs sites 😄

docs/src/components/navigation/table-of-contents.tsx
BarnabyBishop: I went through the same thought process just before committing 😂. Since it is all on-page I don't see any reason that we would need to use Next's link.

docs/src/components/heading.tsx
BarnabyBishop: - [ ] anchor = false

docs/src/components/navigation/table-of-contents.tsx
BarnabyBishop: - [ ] TOC Heading structures

docs/src/app/docs/layout.tsx
BarnabyBishop: Was trying to minimise the merge strife when Johan merges his PR, but you are right, may as well do it properly!
- [ ] Grid to page

docs/src/components/navigation/table-of-contents.tsx
jtbrolin: Yeah I want to remember in EDS docs site, we had some bugginess with the "scroll down to #" when using next/link for the TOC links

ChrisLaneAU (approved):
Lovely 🙌

jtbrolin (approved):
Works like a charm! Nice werk 🙌🏻 🙌🏻

PR #276, opened 2023-05-24
KEY-117: Landing page 2.0

jtbrolin:
Landing page re-styled as per [Figma design](https://www.figma.com/file/BSHTsF2O9iSRiW4ecmh4Bh/Keystatic-%C2%B7-Website?type=design&node-id=740-7409&t=u2p2u4oWuNeNc5qk-4)

Screen Shot 2023-05-25 at 11 26 45 am
Screen Shot 2023-05-25 at 11 26 31 am


docs/public/images/hero-editor.png
emmatown, line 1: Could we please use the real editor at least a screenshot of the real editor here because this is just a design of something that doesn't exist? We made the same mistake with Keystone's site of using aspirational designs that weren't what actually exists (and it never got implemented like that) and I'd like to not make the same mistake again.

BarnabyBishop, line 1: [This PR](https://github.com/Thinkmill/keystatic/pull/277) removes all images from the `public` directory (and the directory itself) to move everything to cloud storage. 

It would be good to move the hero images there, and get that PR when it is merged so the templates component it using the cloud images.

jtbrolin: Was testing the real current editor, and yeah there are some small inconsistencies like padding, the purple badge thingy next to heading, image width potentially, and the word "symbolic" has a weird font-size. I've asked Tuan for the source to that image so we can get is closer to the real thing!

FYI implementing the _actual_ editor in the Landing Page is planned for future I think 👍🏻

ChrisLaneAU: @jtbrolin if you need to delete some images from cloud-storage, let me know, I'm happy to do it for you (you can't overwrite/delete images from the UI).

jtbrolin: Ahh gotcha!! Will do, and thanks ❤️

docs/src/components/templates.tsx
```diff
@@ -302,18 +298,17 @@ function CtaButtons({
   // Available template
   // ------------------------------
   return (
-    <div className="mt-6 flex gap-3">
+    <div className="mt-6 flex gap-4">
       <Button className="flex-1" onClick={onClick}>
         Get started
```
ChrisLaneAU, line 303: What do you think about extending what's announced with these buttons? "Get started" doesn't have any context on its own. Maybe we could just add an `aria-label` to each, e.g. this one could be:

```
aria-label={`Get started with a ${id === 'blank' ? 'blank' : name} template`}
```

docs/src/components/features.tsx
```diff
@@ -0,0 +1,53 @@
...
+        {features.map(feature => (
+          <li
+            key={feature}
+            className="flex items-center gap-2 bg-keystatic-gray rounded-lg py-3 px-4 md:py-4 md:px-6 shrink-0 first:ml-auto last:mr-auto"
+          >
+            <svg
+              className="w-3 md:w-4 shrink-0"
+              xmlns="http://www.w3.org/2000/svg"
+              fill="none"
+              viewBox="0 0 17 19"
```
ChrisLaneAU, line 37: This image should also be `aria-hidden="true"`, same as the one above. I think this should apply to all of the svgs on the home page, IMO.

As a side note - it would be nice to move all of these svgs to their own components as it would help the readability of these files and we could blanket apply `aria-hidden="true"` with an override prop.

docs/src/components/features.tsx
```diff
@@ -0,0 +1,53 @@
...
+          />
+        </svg>
+      </div>
+
+      <ul className="px-6 py-8 overflow-x-auto flex gap-4 md:gap-8">
```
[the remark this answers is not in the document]
ChrisLaneAU: Also, lists don't get announced as lists in VoiceOver unless you add `role="list"` 🫠

Are you able to add that here, and in any other lists on the home page?

docs/public/images/hero-editor.png
jtbrolin: Hey @emmatown 👋🏻 

Here's a proposed new image. I have adjusted so it's closer to the real thing, although not 1:1 because of reasons (x-axis padding in Admin UI is inline with other inputs padding which looks good in context, but looks a bit shit on a marketing site on its own, and text color is slightly more subtle so as to not take too much focus to the content). 
What you think?
hero-editor-current

The _actual_ real thing that I've compared with:
Screen Shot 2023-05-25 at 1 43 47 pm

emmatown: Let's just do the real thing, POC: https://github.com/Thinkmill/keystatic/commit/eb554f6aef2534d1a8b8696eb58ca24c05d297d4

docs/src/components/features.tsx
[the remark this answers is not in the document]
jtbrolin: > confirming that we want these to be horizontally scrolled on mobile

Yeah this is how I interpreted the static figma designs, so it's intentional, but could very well change as we go if peeps think it's no good :) 

> Also, lists don't get announced as lists in VoiceOver

Ahh didn't know that! Thaanks :)

- [ ] Add role="list" to any list on homepage

docs/src/components/features.tsx
jtbrolin: - [ ] aria-hidden="true" on all svg's on landing page

docs/src/components/hero.tsx
```diff
@@ -69,28 +68,42 @@ export default function Hero() {
...
-          ))}
-        </ul>
+
+        <p className="text-sm text-neutral-600 text-center">
+          Live edit content on GitHub or your local file system, without
+          disrupting your existing code and workflows.
+        </p>
+
+        <div className="relative w-full h-[100vw] md:h-[22vw] lg:h-[25vw] xl:h-[20rem]">
+          <Image
```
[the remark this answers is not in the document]
jtbrolin: Nice one!! I saw some console warning about this I think, but have switched to a plain  for now and will change that to use BB's CloudImage component from the Assets PR when that gets in I was thinking 👍🏻

docs/src/components/features.tsx
jtbrolin: @ChrisLaneAU I've put some <svg>'s in icon components now, the ones I thought made sense, and gets it a little bit more dry, but let me know what you think if I can improve more 🙌🏻 Cheers

docs/public/images/hero-editor.png
BarnabyBishop: Awesome, thanks for the POC @emmatown - @ChrisLaneAU  will be looking into that in a new ticket/PR

docs/src/components/features.tsx
ChrisLaneAU: Love it, that's perfect. Thanks for doing this :)

ChrisLaneAU (approved):

PR #277, opened 2023-05-25
Docs - assets

BarnabyBishop:
- Moved all images from notion to cloud and added in content
- Moved landing page images to cloud 
- Removed fonts and implemented `next/font/google`
- Updated metadata to match current Next metadata API
- Moved favicon/icons to `app` directory to use next's new [icon functionality](https://nextjs.org/docs/app/api-reference/file-conventions/metadata/app-icons)
- Tidied up some content


docs/src/content/pages/installation-next-js.mdoc
ChrisLaneAU, line 1: We've still got some `Missing image`s in this file - do we need to fix them now?

docs/src/content/pages/automated-setup-with-template.mdoc
```diff
@@ -20,7 +20,8 @@ You _can_ do that manually… but we’ve created a little semi-automated onboar
 
 You’ll be promoted to go through a few steps. You can read the **step by step walkthrough** below for details, but here’s a visual representation of the _mostly automated_ journey you’ll be taken on:
```
ChrisLaneAU, line 21: Small typo - `promoted` -> `prompted`

docs/next.config.js
```diff
@@ -6,4 +6,14 @@ module.exports = {
     appDir: true,
     externalDir: true,
   },
+  images: {
```
ChrisLaneAU, line 9: Maybe not for solving now as this is likely part of a cloud-image refactor but using `next/image` out of the box with `keystatic.io/images` feels like a clash of technologies. i.e. two competing image optimisation services + CDNs. We could turn off Next's image optimisation here with [unoptimized](https://nextjs.org/docs/app/api-reference/components/image#unoptimized) or a [custom loader](https://nextjs.org/docs/app/api-reference/next-config-js/images). Or maybe it's fine taking the path we have with creating our own image component.

docs/src/content/pages/automated-setup-with-template.mdoc
BarnabyBishop: Nice pick up, fixed.

docs/src/content/pages/installation-next-js.mdoc
BarnabyBishop: Yes! I've used the images that are still relevant from the astro docs, but will want to replace all of them with Next.js eventually. This will do for now

docs/next.config.js
BarnabyBishop: Yeah I switched back and forward with this one, but you are right we shouldn't be using next/image here. I'll replace with our own images component.

ChrisLaneAU: Cool cool. There's more to solve WRT layout shift and figuring out + providing height/width to our cloud-image's `img` element. It would be nice to have a clear picture of how `next/image` could be best used by other folks picking up keystatic cloud as well, if they were keen to stick with next/image. Since we control storage, CDN and optimising, rather than just optimising, to me it would make sense to streamline the whole flow. Anywho, just thinking out load ATM 😊

docs/next.config.js
```diff
@@ -6,4 +6,14 @@ module.exports = {
     appDir: true,
```
ChrisLaneAU, line 6: I was just reading that this is no longer needed https://nextjs.org/docs/app/api-reference/next-config-js/appDir

BarnabyBishop: Nice!

docs/src/components/document-renderer.tsx
```diff
@@ -71,7 +71,7 @@ const getRenderers = (
       }
       return (
         <code
-          className="[&>pre]:p-4 [&>pre]:rounded-md text-sm"
+          className="[&>pre]:whitespace-break-spaces [&>pre]:break-all [&>pre]:p-4 [&>pre]:rounded-md text-sm"
```
jtbrolin, line 74: Niice :)

jtbrolin (approved):

PR #295, opened 2023-06-01
KEY-134: Coming soon

jtbrolin:
- Adds an option for a Coming soon (no link) option to the docs nav as per https://linear.app/keystatic/issue/KEY-134/sort-out-remix-menu-item


docs/src/utils/reader.ts
```diff
@@ -15,11 +15,18 @@ export async function getNavigationMap() {
       const { discriminant, value } = link;
       const page = discriminant === 'page' && value ? pagesBySlug[value] : null;
       const url = discriminant === 'url' ? value : `/docs/${page?.slug}`;
+      const comingSoon = discriminant === 'coming-soon' ? true : false;
+      const comingSoonBadgeLabel =
```
BarnabyBishop, line 19: Is there a case where coming soon is `true` but there is no `comingSoonBadgeLabel`?
If not we could probably just have the `comingSoonBadgeLabel` variable to pass down as props everywhere, and use that as the test case.

BarnabyBishop (commented):
Nice one! One small question but all good otherwise!

docs/src/utils/reader.ts
jtbrolin: > Is there a case where coming soon is true but there is no comingSoonBadgeLabel?

Probably not 😅 That was more me playing with "how would I do that if I wanted to" so totally happy to do whichever

Thinking about the use case we might not even need to customise the badge label, we could just decide that it should always be "Soon" (is that good btw?).. Could really just pass down a comingSoon boolean if that option was selected.. What you reckon? @BarnabyBishop

BarnabyBishop: > Thinking about the use case we might not even need to customise the badge label
Since this is temporary, and only for one link, you are right we probably don't need the custom label.

jtbrolin: Have removed that conditional field now and simplified 👍🏻 🚚  @BarnabyBishop

docs/src/utils/reader.ts
```diff
@@ -15,11 +15,13 @@ export async function getNavigationMap() {
       const { discriminant, value } = link;
       const page = discriminant === 'page' && value ? pagesBySlug[value] : null;
       const url = discriminant === 'url' ? value : `/docs/${page?.slug}`;
+      const comingSoon = discriminant === 'coming-soon' ? true : false;
```
ChrisLaneAU, line 18: You don't need the ternary part of this. `discriminant === 'coming-soon'` evaluates to `true` or `false`. As a general rule, if you find yourself adding a boolean literal in a ternary, it usually means you can simplify the code.

Also, because this variable isn't reused, I would suggest removing this line then making line 24: `comingSoon: discriminant === 'coming-soon'`.

jtbrolin: Ahh silly me! Thanks Laney, I was a bit quick on the gun there 😅

BarnabyBishop (approved):
Coming soon? more like coming now! 🚚

docs/src/components/navigation/nav-item.tsx
```diff
@@ -22,23 +30,42 @@ export function NavItem({ label, href, level, title, tabIndex }: NavItemProps) {
...
+        <div
+          className={cx(
+            styleShared,
+            'inline-flex gap-1 items-baseline text-neutral-500'
+          )}
         >
           {label}
-        </a>
-      </Link>
+          {comingSoon && (
```
ChrisLaneAU, line 48: You don't need this check since you're doing it on line 40. Sorry, I missed this earlier.

ChrisLaneAU (approved):

ChrisLaneAU (approved):

docs/src/components/navigation/nav-item.tsx
jtbrolin: Ahh thank you 😅🙈

PR #299, opened 2023-06-05
KEY-135: 404 page

jtbrolin:
- Updates repo to next 13.4.4 (next link bug related to not-found fixed)
- Implements the `notFound()` for docs pages
- Uses the new `app/not-found.tsx` for 404's
-- Added back a RootLayout at top level to support the not-found.tsx
- Minor style tweaks to Thank you and Privacy policy pages

NOTE:
Nav items still doesn't appear on deployment on /docs/something 404 page (hamburger menu on small screens)


BarnabyBishop:
> Nav items still broken on deployment on 404 page (hamburger menu on small screens)

As we discussed, this is only an issue on 404s in the dynamic `/docs/[slug]` route. And I wouldn't call it broken, it just doesn't display all the nav items, just the static ones.

Since the site is now mostly statically generated, Next/Vercel seems good at caching 404s:

- `/notapage` renders the statically generated `not-found` with menu
- `/nota/subpage` renders the statically generated `not-found` with menu
- Even `/docs/notapage/notasubpage` renders the statically generated `not-found` with menu

So because vercel/next knows these pages can't exist, it will give it the statically generated `not-found` page. 

However even though we specify the paths for the `/docs/[slug]` route with `generateStaticParams`, Next/Vercel is still sending it to the server to try and render any page that isn't statically generated. I'm not sure if this is:
- A vercel thing due to [ISR](https://nextjs.org/docs/pages/building-your-application/data-fetching/incremental-static-regeneration)
- A next.js config thing
- A bug in Next13

I had a quick scan of Next issues but nothing stood out, any I'm not sure it is worth spending much more time on. Knowing our luck no-one will ever click that menu if they get a 404 in the docs root directory, they will just head to the homepage 😅

docs/src/app/keystatic/layout.tsx
```diff
@@ -1,12 +1,5 @@
...
 export default function RootLayout() {
-  return (
-    <html>
-      <head />
-      <body>
-        <KeystaticApp />
-      </body>
-    </html>
-  );
+  return <KeystaticApp />;
```
BarnabyBishop, line 4: The only thing to note here is that the Keystatic Admin UI will get the root layout with:

- metadata - I think this is fine, we can overwrite it in future if we need to
- Analytics - This might actually be good for when we set it up properly and encourage users to check our the keystatic admin for the keystatic page!

This is just a FYI, no changes requested

BarnabyBishop (approved):
Nice!

ChrisLaneAU (approved):

PR #300, opened 2023-06-06
Docs - introduction & quick start

BarnabyBishop:
- Added a simple Introduction page to get us started
- Added a simple quick start guide (this will be updated when we have a CLI)
- Added a tags component block
- Added a custom layout renderer for ✨ styles ✨


docs/next.config.js
```diff
@@ -5,4 +5,13 @@ module.exports = {
   experimental: {
     externalDir: true,
   },
+  async redirects() {
+    return [
+      {
+        source: '/docs',
+        destination: '/docs/introduction',
```
ChrisLaneAU, line 12: Should we introduce a `routes` object and type? Something like:

```ts
// routes.ts
export const routes = {
  home: '/',
  docs: '/docs',
  'docs-introduction': '/docs/introduction',
} as const;

export type Routes = typeof routes;
```

docs/src/components/document-renderer.tsx
```diff
@@ -108,6 +108,20 @@ const getRenderers = (
     divider: () => {
       return <hr className="border-keystatic-gray my-2" />;
     },
+    layout: ({ layout, children }) => (
+      <div
+        className="grid gap-4"
+        style={{
+          gridTemplateColumns: layout.map(x => `${x}fr`).join(' '),
```
ChrisLaneAU, line 115: Do you mind if we avoid single letter variables? I find them hard to read 😅

docs/next.config.js
BarnabyBishop: It isn't super clear to me what benefit this will bring? Perhaps if we start referencing routes elsewhere? But until that time I'd be inclined to keep it simple.

docs/src/components/document-renderer.tsx
BarnabyBishop: Fair call, I did a cheeky copy/paste from the default document-render layout component in core for this one to add our tailwind classes 🙈

What name would make sense? `colSpan` is the best I can come up with but I'm not sure that is actually correct 🤔

BarnabyBishop: `layoutColumns`, or `columns` perhaps, not exactly what it is but close. It is like a column index with possible span information 😅

ChrisLaneAU: I would go `columnSize` or `columnFraction`.

jtbrolin: columnFractions? 😅

jtbrolin: layoutFraction in singular would make sense to me actually, but doesn't really matter :D

BarnabyBishop: The people have spoken

jtbrolin (approved):
Awesome work getting this first version together @BarnabyBishop !!

I had some tiny things I wanted to tweak, but was easier to just work them out for myself in code, so I branched off and pushed my tweaks here https://github.com/Thinkmill/keystatic/tree/docs-introduction-style-tweaks

- Some spacing around `<ul>`'s
- Smaller font-size in the layout-area
- Added some dividers to make easier to digest and clarifies what section the links belong to
- Link text: I thought it's worth mentioning the page name Quick start / Automated setup e.g in the link, and removed "here"'s

Do as you wish with that branch :) And let me know if u need me to make a PR or add tweaks as comments instead :)

BarnabyBishop:
> I had some tiny things I wanted to tweak, but was easier to just work them out for myself in code, so I branched off and pushed my tweaks here https://github.com/Thinkmill/keystatic/tree/docs-introduction-style-tweaks

Amazing. I was thinking of getting you just to make any style/copy changes direct in this branch 😄 

I've merged your changes in. 🪄

docs/src/components/document-renderer.tsx
```diff
@@ -90,24 +90,45 @@ const getRenderers = (
...
+        {children.map((element, i) => (
+          <div
+            key={i}
+            className="rounded-lg bg-keystatic-gray-light p-4 text-sm"
+          >
+            {element}
+          </div>
+        ))}
+      </div>
+    ),
```
jtbrolin, line 131: @BarnabyBishop I forgot mobile in my tweaks 📱 :) This code seems to work on my local, but you might be able to improve:

```suggestion
    layout: ({ layout, children }) => {
      const layoutStyle = `sm:grid-cols-[${layout
        .map(layoutFraction => `${layoutFraction}fr`)
        .join('_')}]`;

      return (
        <div className={`grid gap-6 my-2 grid-cols-1 ${layoutStyle}`}>
          {children.map((element, i) => (
            <div
              key={i}
              className="rounded-lg bg-keystatic-gray-light p-4 text-sm"
            >
              {element}
            </div>
          ))}
        </div>
      );
    },
```

If we add more layout options in the config then this would have to be smarter I suppose..

BarnabyBishop: Ack mobile, I parked it as I couldn't think of a solution and never came back to it. 

Is your example working on desktop? For me the `sm`/desktop styles aren't working. 

My understanding of tailwind is any of these dynamic styles in square brackets, e.g. `grid-cols-[1fr 1fr]` need to be "hardcoded" (for lack of a better word) in the className as tailwind will find these at build time and create custom css classes for it. 

If we make it dynamic like here those styles aren't known until run-time (when the app is running) so tailwind can't create these custom classes. So I don't think this will work.

We also can't just use the `style` attribute as that doesn't support media queries to my knowledge. I'll have to think of a solution here 🤔

jtbrolin: Hmm maybe when I tried this code I had tried hardcoded first so it was cached or something when it worked for me 🤔  

All good, I'd also be fine just hardcoding `grid-cols-1 sm:grid-cols-2` for now, or make a custom component block instead potentially, but I wonder what the recommended solution for the document field's Layout when it comes to responsiveness would be, or are they removing that (have a subtle memory of reading something somewhere but might be wrong)

BarnabyBishop: > I'd also be fine just hardcoding grid-cols-1 sm:grid-cols-2 for now

This is a good call as in our document config we only support two column layouts anyway.

> I wonder what the recommended solution for the document field's Layout when it comes to responsiveness would be

Use emotion instead of tailwind? 😁

PR #301, opened 2023-06-06
Docs: Move the field content into a 'fields' directory/path

BarnabyBishop:
This PR moves our field content into a folder (and url path) called `/fields`


docs/src/app/(public)/docs/[...slug]/page.tsx
```diff
@@ -29,6 +30,7 @@ export default async function Docs({ params }: { params: { slug: string } }) {
           {page.title}
         </h1>
         <div className="flex flex-col gap-4">
+          {/* @ts-expect-error react-server component */}
```
Comment (BarnabyBishop) [inline docs/src/app/(public)/docs/[...slug]/page.tsx:33]: I'm a bit bamboozled why this is suddenly throwing a TS error:

image

Will do some digging, but don't _think_ it should be a blocker

Comment (ChrisLaneAU) [reply docs/src/app/(public)/docs/[...slug]/page.tsx]: Possibly due to the updated react types where promises are valid?

Comment (BarnabyBishop) [reply docs/src/app/(public)/docs/[...slug]/page.tsx]: > Possibly due to the updated react types where promises are valid?

Good call, my local was out of date.

jtbrolin:
@BarnabyBishop I just realised /docs/fields will be a 404, is that something we want to redirect as well or do we consider that fine? Unsure of common practices in this area, when not having landing pages

BarnabyBishop:
> @BarnabyBishop I just realised /docs/fields will be a 404, is that something we want to redirect as well or do we consider that fine? Unsure of common practices in this area, when not having landing pages

I think its fine. There isn't anywhere sensible to redirect it to anyways.

PR #306, opened 2023-06-09
KEY-148: Fields API content updates

jtbrolin:
- Replaces the field type signatures in code to just link out to docsmill
- Fields copy tweaks
- Renamed Multi-Select page -> multiselect (without the hyphen for consistency with field api naming)


jtbrolin:
Note, these things from the ticket will be added in another PR:
- Add the slugField option to fields that support it, and mention it in the slug field itself
- Render each field on the field page for a live example. Since the site doesn't have a dark mode see if we can force light-mode

docs/src/content/pages/fields/object.mdoc
```diff
@@ -1,11 +1,13 @@
 ---
 title: Object field
 ---
-The `object` field lets you structure complex object schemas that can contain any other field types.
+The `object` field is used to create complex object schemas that can contain any other fields. 
 
-## Usage example
+It's particularly useful in conjunction with the [array](/docs/fields/array) field (that only accepts one single field) when you need multiple fields for each item in the array.
```
BarnabyBishop, line 6: It is also useful for conditional and blocks fields, should we add those examples there too, or is it getting to wordy?

ChrisLaneAU, line 6:
```suggestion
It's particularly useful in conjunction with the [array field](/docs/fields/array) (that only accepts one single field) when you need multiple fields for each item in the array.
```

docs/src/content/pages/fields/relationship.mdoc
```diff
@@ -3,7 +3,7 @@ title: Relationship field
 ---
 The `relationship` field is a reference to the `slug` of a specific collection entry.
 
-To create a one-to-many relationship, wrap the `relationship` field inside an [array field](/docs/array-field).
+To create a one-to-many relationship, wrap the `relationship` field inside an [array](/docs/fields/array) field.
```
ChrisLaneAU, line 6:
```suggestion
To create a one-to-many relationship, wrap the `relationship` field inside an [array field](/docs/fields/array).
```

ChrisLaneAU (commented):
Looking great! These explanations and up to date refs are nicely practical! The only comment (which I've repeated a bunch of times 😅) is increasing the context of the links.

docs/src/content/pages/fields/child.mdoc
```diff
@@ -3,7 +3,7 @@ title: Child field
 ---
 The `child` field allows you to embed an editable region inside of a component block preview.
 
-See the [document field](/docs/document-field) for more information about component blocks.
+See the [document](/docs/fields/document) field for more information about component blocks.
```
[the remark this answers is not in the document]
jtbrolin: Ahh yes great point! Have fixed!

docs/src/content/pages/fields/object.mdoc
jtbrolin: Have tweaked now, let me know what you think!

docs/src/content/pages/fields/blocks.mdoc
```diff
@@ -0,0 +1,31 @@
+---
+title: Blocks field
+---
+The `blocks` field is similar to the [array field](/docs/fields/array) in that you can create "Add one more" scenarios, but with the difference that it lets you define a separate field schema for each instance.
```
jtbrolin, line 4: Added this now since last review :) Please advise if you can think of improvements 🙏🏻

docs/src/content/pages/fields/image.mdoc
```diff
@@ -1,11 +1,15 @@
 ---
 title: Image field
 ---
-The `image` field is used to upload an image.
+The `image` field is used to store an image. In the Admin UI it renders a button to choose an image file. When a file has been selected it shows a preview along with a button to remove.
```
BarnabyBishop, line 4: Sorry I missed this last review, but this second sentence may be too detailed? The Image field UI is something that I can foresee changing (e.g. to make a nice dropzone) so the description has a higher chance of getting out of date than other simpler fields. 

Perhaps simplify it? If we get the fields integrated the user will see what it looks like anyways.

```suggestion
The `image` field is used to store an image. In the Admin UI it renders an image picker component.
```

BarnabyBishop (approved):
Once tiny suggestion otherwise lookin' good. 🤏

ChrisLaneAU (approved):

PR #307, opened 2023-06-12
@keystatic/create

BarnabyBishop:
This package has the MVP of a create-keystatic-app style CLI.

For now it simply creates a new Next.js app with Keystatic but in future can have more template options, and potentially the ability to add Keystatic to an existing app.

**Initial template**
I have stored the initial blank template in a `templates` folder. I chose to put it in the monorepo rather than a seperate repo so it keeps up with Keystatic's versioning. @emmatown - it would be great to get your thoughts on this.

Rather than bundle templates with the CLI package it uses [gitget](https://www.npmjs.com/package/giget) to download the templates from github. We could use this in future to download templates we may not want in the monorepo such as the blog, marketing website templates etc (or would we want these in the monorepo?)

**Testing**
You can test this locally by running `pnpm dev:create`. In this branch the CLI will fail as it can't find the template as it isn't in `main` yet. 

To test it fully open `packages/create/src/actions/create-project.ts` and add `#create-cli` to the end of the template path on  line 9.


emmatown (commented):
This looks great!! Definitely agree with having them in the repo (and not in the CLI) though one concern: if we e.g. have a breaking change and need to update the templates, we only want the updated templates to be used _after we release the breaking change_. maybe we could make the templates literally npm packages that we version and just download them from the npm registry in the CLI?

packages/create/src/actions/project-name.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  if (isCancel(projectName)) cancelStep();
+
+  if (typeof projectName === 'string') {
+    ctx.projectName = projectName;
+
+    if (projectName === '.' || projectName === './') {
+      ctx.cwd = process.cwd();
+    } else {
+      ctx.cwd = projectName;
```
ChrisLaneAU, line 22: Does this need any prefixing? Like `./` or `process.cwd()`?

packages/create/src/actions/project-name.ts
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    const isDirectoryEmpty =
```
ChrisLaneAU, line 25: Should this be `isDirectoryNotEmpty` or `const isDirectoryEmpty = !(...);`? It seems to be truthy when the directory has files.

templates/next-app-dir/app/page.tsx
```diff
@@ -0,0 +1,27 @@
...
+      <h1>Keystatic ⚡️</h1>
+      <p>This homepage shows how to load a collection from the reader API.</p>
+      <p>
+        <a href="/keystatic">Click here to visit the Admin UI</a>, or the link
+        below to view a post in the collection.
+      </p>
+      <h2>Posts</h2>
+      <ul>
+        {posts.map(post => (
+          <li>
```
ChrisLaneAU, line 18: Should we put a `key` on here? i.e. `key={post.slug}`?

ChrisLaneAU (approved):

templates/next-app-dir/app/page.tsx
BarnabyBishop: 💯

packages/create/src/actions/project-name.ts
BarnabyBishop: It seems to work without it. It also allows you to put in an absolute path, e.g. `/Users/barney/projects/ks-test` which is handy for testing.

packages/create/src/actions/project-name.ts
BarnabyBishop: haha yes good catch

ChrisLaneAU (approved):

BarnabyBishop:
> maybe we could make the templates literally npm packages that we version and just download them from the npm registry in the CLI?

Great point. I've updated the template downloader to download from npm directly. I haven't added the major version checker we discussed yesterday but that will be easy to add once we have these initial versions published.

Should we move the templates from the `templates` directory to the `packages` directory since they are now packages, or keep them seperate? 

For anyone testing, with this new update we can't test the whole CLI with template flow with the actual template until it is published.

However you can test the CLI/downloading template flow by updating this line in `packages/create/src/actions/create-project.ts` to point to a published npm package (`@keystatic/next` works fine):
```
const templates = {
  nextjs: '@keystatic/templates-nextjs',
};
```

You can test the actual template by copying the files in `templates/nextjs` out of the monorepo and `pnpm i && pnpm dev`

templates/nextjs/app/page.tsx
```diff
@@ -0,0 +1,27 @@
...
+      <p>This homepage shows how to load a collection from the reader API.</p>
+      <p>
+        <a href="/keystatic">Click here to visit the Admin UI</a>, or the link
+        below to view a post in the collection.
+      </p>
+      <h2>Posts</h2>
+      <ul>
+        {posts.map(post => (
+          <li key={post.slug}>
+            <a href={`/${post.slug}`}>{post.entry.title}</a>
```
jtbrolin, line 19: Could these links be next/links or is there intentions behind? :)

jtbrolin (approved):
So cool!! Have tested according to your last comment and everything's looking great to me, keen to try it from npm when published!

Just left a tiny question, but approved!

emmatown:
> Should we move the templates from the templates directory to the packages directory since they are now packages, or keep them seperate?

I'd keep them seperate

emmatown (dismissed):

templates/nextjs/app/styles.tsx
emmatown, line 1: Why not put this in an actual css file?

templates/nextjs/app/[slug]/page.tsx
```diff
@@ -0,0 +1,19 @@
...
+export default async function Post({ params }: { params: { slug: string } }) {
+  const { slug } = params;
+
+  const post = await reader.collections.posts.read(slug);
+
+  if (!post) return <div>Post not found!</div>;
+
+  return (
+    <div>
+      <h1>{post?.title}</h1>
```
Comment (emmatown) [inline templates/nextjs/app/[slug]/page.tsx:13]:
```suggestion
      <h1>{post.title}</h1>
```

emmatown:
(okay, after d30fe25, I think it'll probably work

BarnabyBishop:
> (okay, after [d30fe25](https://github.com/Thinkmill/keystatic/commit/d30fe25968cf21e3b86be01b8fbe938aef9b1702), I think it'll probably work)

 🤦‍♂️ missed the most important part. Cheers 👍

templates/nextjs/app/page.tsx
BarnabyBishop: No intention! Updated

templates/nextjs/app/styles.tsx
BarnabyBishop: Yeah I dunno what I was thinking there. Moved.

ChrisLaneAU (approved):

BarnabyBishop:
FYI I'm still getting to the bottom of one issue. When trying to `next build` the nextjs template after installing through the CLI I'm getting this error:

```
./node_modules/.pnpm/@types+node@16.11.13/node_modules/@types/node/globals.d.ts:72:13
Type error: Subsequent variable declarations must have the same type.  Variable 'AbortSignal' must be of type '{ new (): AbortSignal; prototype: AbortSignal; abort(reason?: any): AbortSignal; timeout(milliseconds: number): AbortSignal; }', but here has type '{ new (): AbortSignal; prototype: AbortSignal; }'.
```

BarnabyBishop:
> FYI I'm still getting to the bottom of one issue.

Fixed this by making the tsconfig of the template a bit looser, since we don't want to assume users want to write typescript, and to make it more similar to Next's tsconfig which it will annoyingly update for us anyways...

Also fixed a few other small things blocking the ability to directly deploy the template to vercel

PR #317, opened 2023-06-14
Change aside component in docs site

emmatown:
Right now, this will be a worse editor experience but it's how the Markdoc should look and will be necessary for the new editor (I want to try the new editor on the docs content to see how it is)


BarnabyBishop (approved):
This markdoc is much nicer. 

Looking forward to seeing it in the new editor ui 👍

ChrisLaneAU (approved):
🙌

PR #318, opened 2023-06-15
Break long words in links

### LINKED ISSUE #314: Docs: horizontal overflow
Links don't wrap, causing page to overflow horizontally

simonswiss:
Fixes #314

I make use of an arbitrary variant here, but this would only apply the `word-break: break-all` CSS to `a` children tags inside the `DocumentRenderer`. 

Are there potentially other scenarios that should be covered?

If not, this should be good to go.


jtbrolin (approved):
Noice!

florian-lefebvre (approved):

BarnabyBishop (approved):

PR #320, opened 2023-06-15
Manual GitHub setup guide

simonswiss:
I know I myself requested we depart from screenshot-heavy guides, but... 

... it's pretty hard to do a guide about going through specific screens and UIs without providing screenshots 😅


emmatown (commented):
naming nit: could we call this like "local GitHub app setup" or something and the other one like "Vercel GitHub app setup"? manual vs automatic is quite inaccurate since they're both quite "automatic". from my perspective, "manual" setup means clicking around in GitHub's UI to create the GitHub app

simonswiss:
This guide is meant to eventually also show how to setup the remote (not necessarily Vercel) by copying the environment variables over to the host. 

But I do agree, this doesn't really feel any more "manual" — you still go through the automated GitHub app creation, just not via a starter template. I think your naming suggestion makes sense — I could make those changes part of this PR. 

What are your thoughts @BarnabyBishop?

BarnabyBishop (approved):
Looking good

PR #323, opened 2023-06-15
KEY-148: Field demos

jtbrolin:
- Adds interactive field demos to the Fields API pages (note: only the ones that were cheap to add at this stage)
- Adds info about `slugField` to `/docs/fields/slug`, `/docs/fields/array`, and `/docs/collections-and-singletons`

NOTE: 
- I also had to move some things around to make the Voussoir styles work correctly:
-- The font is now implemented as per https://nextjs.org/docs/app/building-your-application/optimizing/fonts#with-tailwind-css
-- The @tailwind/forms global styles were clashing with Voussoir, so now only importing the class names which needs to be used on all non-Voussoir inputs on docs site


docs/src/components/document-renderer.tsx
```diff
@@ -120,9 +129,9 @@ const getRenderers = (
     layout: ({ children }) => {
       return (
         <div className="grid gap-6 my-2 grid-cols-1 sm:grid-cols-2">
-          {children.map((element, i) => (
+          {children.map((element, index) => (
             <div
-              key={i}
+              key={index}
```
jtbrolin, line 134: Changed for consistency w rest of file

docs/src/components/document-renderer.tsx
```diff
@@ -97,7 +106,7 @@ const getRenderers = (
         return (
           <ol className="text-keystatic-gray-dark list-decimal list-inside mt-2">
             {children.map((child, index) => (
-              <li key={index} className="mb-2 leading-tight">
+              <li key={index} className="mb-2">
```
jtbrolin, line 109: Regretted adding this cause it made the lines too tight and hard to read

docs/src/components/document-renderer.tsx
```diff
@@ -80,7 +89,7 @@ const getRenderers = (
 
       return (
         <div
-          className="[&>pre]:whitespace-break-spaces [&>pre]:break-all [&>pre]:p-4 [&>pre]:rounded-lg [&>pre]:border [&>pre]:border-keystatic-gray [&>pre]:bg-keystatic-gray-light text-sm my-2"
+          className="[&>pre]:whitespace-break-spaces [&>pre]:break-all [&>pre]:px-6 [&>pre]:py-4 [&>pre]:rounded-lg [&>pre]:border [&>pre]:border-keystatic-gray [&>pre]:bg-white text-sm my-2"
```
jtbrolin, line 92: Change code examples to have white bg, cause I felt it looks cleaner/fresher especially now that the visual field demos are added and have a light grey bg as well

docs/src/components/fields/integer.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+  return (
+    <FieldDemoFrame>
+      <IntegerFieldInput
+        label="Age"
+        onChange={setValue}
+        autoFocus={false}
+        forceValidation={false}
+        description="The person's age"
+        validation={{ min: 0, max: 120 }}
```
ChrisLaneAU, line 18: We'll need to update this for [Maria](https://en.wikipedia.org/wiki/Maria_Branyas) in 3.5 years 😉

docs/src/components/fields/multiselect.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+import { FieldDemoFrame } from './frame';
+import { MultiselectFieldInput } from '../../../../packages/keystatic/src/form/fields/multiselect/ui';
+import { useState } from 'react';
+
+export const MultiselectFieldDemo = () => {
+  const [value, setValue] = useState<readonly string[]>([
+    'surfing',
+    'basketball',
+    'music',
```
ChrisLaneAU, line 11: And `'chess'`?

ChrisLaneAU: Or what about moving the `options` up into an `options` const and then using grabbing the type from that? Something like this:

```
const options = [
  { label: 'Surfing', value: 'surfing' },
  { label: 'Basketball', value: 'basketball' },
  { label: 'Music', value: 'music' },
  { label: 'Chess', value: 'chess' },
] as const;

type Options = (typeof options)[number]['value'];
```

docs/src/content/pages/fields/array.mdoc
```diff
@@ -51,6 +51,34 @@ complexArray: fields.array(
 ),
 ```
 
+## Slug field
+
+The array field's `slugField` option is useful to replace indexes normally in file paths for images or documents e.g.
```
ChrisLaneAU, line 56: Is there an example missing at the end?

docs/src/content/pages/fields/file.mdoc
```diff
@@ -0,0 +1,30 @@
+---
+title: File field
+---
+{% fieldDemo field="file" /%}
+
+The `file` field is used to store a file. In the Admin UI it renders an file picker component.
```
ChrisLaneAU, line 6: Small typo

```suggestion
The `file` field is used to store a file. In the Admin UI it renders a file picker component.
```

docs/src/content/pages/fields/file.mdoc
```diff
@@ -0,0 +1,30 @@
...
+
+In [local storage mode](/docs/local-vs-git-hub-storage-strategy#local-strategy) it will save it as a file in your repository. It [GitHub storage mode](/docs/local-vs-git-hub-storage-strategy#git-hub-strategy) it will commit it into the repository.
```
ChrisLaneAU, line 8: What do you think about some slight copy changes for readability?

```suggestion
In [local storage mode](/docs/local-vs-git-hub-storage-strategy#local-strategy) the file will be saved in your repository. In [GitHub storage mode](/docs/local-vs-git-hub-storage-strategy#git-hub-strategy) the file will be committed into the repository.
```

docs/src/content/pages/fields/slug.mdoc
```diff
@@ -23,6 +25,17 @@ title: fields.slug({
 })
 ```
 
+{% aside %}
+{% component-inline-prop propPath=["icon"] %}📢{% /component-inline-prop %}
```
ChrisLaneAU, line 29: This has new syntax now, right? Might need to merge `main` and update this?

ChrisLaneAU (commented):
Just some small suggestions but looking great 🙌  I find these demo/examples super useful when reading docs 😃

docs/src/components/fields/integer.tsx
jtbrolin: Haha wow 🤯

docs/src/components/fields/multiselect.tsx
jtbrolin: Ahh yeah this is cool, have applied to the multiselect and the select, was there any other place you thought the typing could be improved @ChrisLaneAU? 🙏🏻

ChrisLaneAU: That looks good to me :)

docs/src/content/pages/fields/slug.mdoc
jtbrolin: Ahh thanks for that catch!!

docs/src/content/pages/fields/array.mdoc
jtbrolin: Nah it was just written badly, I've now changed to `etc` at the end. But unsure if this section makes sense so feel free to suggest improvements if you can think of any 🙌🏻

BarnabyBishop (approved):
Looking good, one small comment on "Resume" for the File field but it isn't a biggy.

docs/src/components/fields/file.tsx
```diff
@@ -0,0 +1,27 @@
...
+    data: Uint8Array;
+    extension: string;
+    filename: string;
+  } | null>(null);
+
+  return (
+    <FieldDemoFrame>
+      <FileFieldInput
+        label="Resume"
+        description="The resume for this user"
```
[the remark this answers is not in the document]
jtbrolin: Ahh yes, I can see how this is confusing 😅 Maybe I could just use `CV` as the term, might be clearer? @BarnabyBishop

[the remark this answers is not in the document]
BarnabyBishop: I think Resume is more common in the US than a CV so maybe keep it as is?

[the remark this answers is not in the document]
jtbrolin: Updated the `description` now @BarnabyBishop let me know if that works 🙏🏻 :)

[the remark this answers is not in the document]
BarnabyBishop: _Very_ clear now. Nice!

PR #345, opened 2023-06-19
KEY-154: CLI content

jtbrolin:
- Adds "Copy command" button to homepage hero
- Adds CLI command and small content to /docs/quick-start page

### DIFF
--- a/docs/src/components/copy-command-button.tsx
+++ b/docs/src/components/copy-command-button.tsx
@@ -8,22 +8,20 @@ export function CopyCommandButton() {
   const copyCommandText = 'npm create @keystatic@latest';
   const [isCopied, setIsCopied] = useState(false);
 
-  const handleCopyCommand = () => {
-    copyTextToClipboard(copyCommandText)
-      .then(() => {
-        setIsCopied(true);
-        setTimeout(() => {
-          setIsCopied(false);
-        }, 1500);
-      })
-      .catch(err => {
-        console.log(err);
-      });
+  const handleCopyCommand = async () => {
+    try {
+      await copyTextToClipboard(copyCommandText);
+      setIsCopied(true);
+      setTimeout(() => {
+        setIsCopied(false);
+      }, 1500);
+    } catch (err) {
+      console.log(err);
+    }
   };
 
   return (
     <Button
-      className="border-keystatic-gray items-center py-0"
       impact="light"
       onClick={handleCopyCommand}
       aria-label="Copy npm command for creating a Keystatic project"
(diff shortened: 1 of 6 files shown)


docs/src/components/copy-command-button.tsx
```diff
@@ -0,0 +1,68 @@
+'use client';
+
+import { useState } from 'react';
+import { copyTextToClipboard } from '../utils';
+import Button from './button';
+
+export function CopyCommandButton() {
+  const copyCommandText = 'npm install @keystatic';
```
BarnabyBishop, line 8: This should be `npm create @keystatic`

BarnabyBishop: Actually make it `npm create @keystatic@latest`

jtbrolin: Ahh can't believe I typed that!

docs/src/components/copy-command-button.tsx
```diff
@@ -0,0 +1,68 @@
...
+      .then(() => {
+        setIsCopied(true);
+        setTimeout(() => {
+          setIsCopied(false);
+        }, 1500);
+      })
+      .catch(err => {
+        console.log(err);
+      });
+  };
```
BarnabyBishop, line 22: We should be able to convert this to async/await like so (I haven't tested this!). Bit easier on the eyes then promises.

```suggestion
  const handleCopyCommand = async () => {
    try {
      await copyTextToClipboard(copyCommandText);
      setIsCopied(true);
      setTimeout(() => {
        setIsCopied(false);
      }, 1500);
    }
    catch(err) {
      console.log(err);
    }
  };
```

docs/src/components/copy-command-button.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  return (
+    <Button
+      className="border-keystatic-gray items-center py-0"
```
BarnabyBishop, line 26: On my screen the CLI command box looks a bit washed out and gets lost in the rest of the page:

CleanShot 2023-06-20 at 12 33 08@2x

Do you think we can make it stand out a bit more? Different background or darker border colour?

Basically I'm asking if you can make it pop. 😄

docs/src/components/copy-command-button.tsx
```diff
@@ -0,0 +1,68 @@
...
+      impact="light"
+      onClick={handleCopyCommand}
+      aria-label="Copy npm command for creating a Keystatic project"
+    >
+      <div className="flex gap-4 font-mono text-sm items-center leading-none">
+        {copyCommandText}
+
+        {isCopied ? (
```
BarnabyBishop, line 34: Love this detail

BarnabyBishop (commented):
Looking good, just some small notes

docs/src/components/copy-command-button.tsx
jtbrolin: Haha yeah I'll just make it the normal secondary Button style then and it'll pop more!

BarnabyBishop (approved):
💥

PR #349, opened 2023-06-20
Docs - server rendering

BarnabyBishop:
It turns out our statically generated homepage was completely client-rendering as we were using Next's `useSearchParams` hook without wrapping it in a Suspense boundary. Bit of a hidden gotcha there. 🦶🔫

This fixes that issue. I also wrapped the homepage's mailing list H2 in an anchor tag to see if this fixes the issue that bought this to our attention: https://github.com/Thinkmill/keystatic/pull/344#discussion_r1234693554

It seems it does, would be great to test!


jtbrolin (approved):
Nice research, such a good catch! 🚚 🙌🏻

docs/src/components/templates.tsx
```diff
@@ -117,6 +117,18 @@ function useLastTruthyValue<T>(val: T): T {
 }
 
 export default function Templates() {
+  /* As the <TemplateList /> component uses useSearchParams it must be wrapped in
+   * a Suspense boundary to prevent the whole tree being rendered client-side only.
+   * See: https://nextjs.org/docs/app/api-reference/functions/use-search-params#static-rendering
```
ChrisLaneAU, line 122: Nice find! TIL 🙌 I really need to up my app dir game.

docs/src/components/mailing-list.tsx
```diff
@@ -22,9 +22,11 @@ export default function MailingList() {
       <div className="mx-auto max-w-5xl px-6 pt-12 pb-16 sm:pt-16 sm:pb-24 lg:pt-24 lg:pb-32">
         <div className="grid items-center justify-center gap-8 lg:gap-16 lg:grid-cols-2 xl:gap-32">
           <div className="text-center lg:text-left">
-            <h2 className="text-3xl font-bold sm:text-4xl">
-              Get Keystatic updates
-            </h2>
+            <a href="#mailing-list">
```
ChrisLaneAU, line 25: > This fixes that issue. I also wrapped the homepage's mailing list H2 in an anchor tag to see if this fixes the issue that bought this to our attention: https://github.com/Thinkmill/keystatic/pull/344#discussion_r1234693554

> It seems it does, would be great to test!

Amazing! I didn't even know this was a thing. I also can't find any documentation on this and it's not in the spec (at least that I can find). I was looking at the spec here: https://html.spec.whatwg.org/multipage/browsing-the-web.html#the-indicated-part-of-the-document

> To find a potential indicated element given a [Document](https://html.spec.whatwg.org/multipage/dom.html#document) document and a string fragment, run these steps:
>
> If there is an element [in the document tree](https://dom.spec.whatwg.org/#in-a-document-tree) whose [root](https://dom.spec.whatwg.org/#concept-tree-root) is document and that has an [ID](https://dom.spec.whatwg.org/#concept-id) equal to fragment, then return the first such element in [tree order](https://dom.spec.whatwg.org/#concept-tree-order).
>
> If there is an [a](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element) element [in the document tree](https://dom.spec.whatwg.org/#in-a-document-tree) whose [root](https://dom.spec.whatwg.org/#concept-tree-root) is document that has a [name](https://html.spec.whatwg.org/multipage/obsolete.html#attr-a-name) attribute whose value is equal to fragment, then return the first such element in [tree order](https://dom.spec.whatwg.org/#concept-tree-order).
>
> Return null.

I'm keen to understand this and I've probably missed something but also just want to confirm that it's not the now available server rendering of this component and the `id` still available in the `section` on line 7 that's made this work?

BarnabyBishop: > I'm keen to understand this and I've probably missed something but also just want to confirm that it's not the now available server rendering of this component and the id still available in the section on line 7 that's made this work?

🤦‍♂️ you are so right. I totally forgot there was still the id on the section. I was wondering how it jumped perfectly to the section not the heading but didn't think about it deep enough. 

Good catch I've removed it and it works locally, will test when it has deployed.

ChrisLaneAU (approved):
Nice work! Great solution 🙌

docs/src/components/mailing-list.tsx
ChrisLaneAU: Ahh all good. Makes sense. I tested on FF, Chrome, Safari, Edge and it works!

docs/src/components/templates.tsx
BarnabyBishop: This is such a foot-gun that hopefully Next throws a warning or something in future to let you know that using this hook will break ssr on your entire tree. 

I only found it by brute forcing a solution

PR #350, opened 2023-06-20
KEY-148: Small content tweaks

jtbrolin:
- Style change (only in rendered README content):
-- Increases margin top and bottom around dividers (hr's) to get a more airy and easy-to-digest feel
-- If a heading comes right after a HR (divider) then the heading doesn't need it's own margin-top, leaning on the divider's new big margins instead

- Replaces some old keystatic.thinkmill.com.au -> keystatic.com links
- Replaces some apostrophes to use `'` instead

- Changes `Keystatic is another Thinkmill project` -> `Keystatic is a Thinkmill project` after feedback and discussion


docs/tailwind.config.js
```diff
@@ -40,6 +41,9 @@ module.exports = {
     },
   },
   plugins: [
+    plugin(function ({ addVariant }) {
+      addVariant('peer-adjacent', ':merge(.peer) + &'); // adds custom support for adjacent sibling combinator
```
simonswiss, line 45: Whoaaa, I had never seen (or used) the `:merge()` directive.

I looked it up on MDN before realising it is a new Tailwind thing 😓 

CleanShot 2023-06-20 at 16 16 20@2x

BarnabyBishop (approved):
On fire today 🔥

docs/tailwind.config.js
jtbrolin: Yeah it was fun to learn a bit about and try out, but I'm still curious why they don't have a thing out-of-the-box with the adjacent sibling combinator (or have I missed something? I googled a lot over the weekend haha). Maybe it's just not that useful/common 🤔

PR #369, opened 2023-06-27
KEY-156: Keystatic Blog

jtbrolin:
- Adds a /blog to keystatic.com
- Adds a first blog post about the Keystatic CLI
- Adds a new component block `<Embed />` for tweet and video embeds
- Removes old placeholder content in `docs/src/app/(public)/docs/page.tsx`
- Changes link component to use next/link instead of `<a>` in `docs/src/components/document-renderer.tsx`
- Some minor spacing tweaks to /docs/[...slug] h1 and footer


docs/src/content/blog/the-fresh-keystatic-cli.mdoc
```diff
@@ -0,0 +1,24 @@
...
+summary: >-
+  Kick-start a Keystatic-integrated project with our new "create" CLI, available
+  via NPM.
+authors:
+  - simon-swiss
+  - johan-brolin
+---
+The freshly baked Keystatic CLI is now published on NPM 🔥
+
+Simply run `npm create @keystatic@latest` in your terminal to try it out!
```
simonswiss, line 14:
```suggestion
You can run `npm create @keystatic@latest` in your terminal to try it out!
```

docs/src/content/pages/quick-start.mdoc
```diff
@@ -1,7 +1,6 @@
 ---
 title: Quick start
 ---
-
 ## Keystatic CLI
```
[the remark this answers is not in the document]
ChrisLaneAU: Mind you, I've noticed that nobody does this, so perhaps folks are used to it 🤷

docs/src/content/blog/the-fresh-keystatic-cli.mdoc
simonswiss: I don't mind it myself, but I've heard many time that one should avoid words like "just", "simply" and things that assume everyone is fluent with terminal stuff.

docs/src/content/blog/the-fresh-keystatic-cli.mdoc
```diff
@@ -0,0 +1,24 @@
+---
+title: The fresh Keystatic CLI
+draft: false
+publishedOn: '2023-06-28'
+summary: >-
+  Kick-start a Keystatic-integrated project with our new "create" CLI, available
+  via NPM.
```
simonswiss, line 7:
```suggestion
  via npm.
```

docs/src/content/blog/the-fresh-keystatic-cli.mdoc
```diff
@@ -0,0 +1,24 @@
...
+authors:
+  - simon-swiss
+  - johan-brolin
+---
+The freshly baked Keystatic CLI is now published on NPM 🔥
```
simonswiss, line 12:
```suggestion
The freshly baked Keystatic CLI is now published on npm 🔥
```

ChrisLaneAU (commented):
Lovely addition 😍

simonswiss (changes requested):
Looking good!

My only suggestion (and this is not blocking the merging of this PR) would be to make the Blog homepage header section left-aligned on mobile.

But also this is a personal preference, so 🤷🏼 

CleanShot 2023-06-27 at 13 53 50@2x

docs/src/content/pages/quick-start.mdoc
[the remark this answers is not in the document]
ChrisLaneAU: E.g. Next and Astro aren't doing it. Maybe leave it for consistency among peers.

docs/keystatic.config.tsx
```diff
@@ -144,6 +144,41 @@ export const componentBlocks = {
...
+};
+
+const formatting = {
+  headingLevels: [2, 3],
+  blockTypes: true,
+  listTypes: true,
+  inlineMarks: true,
+} as const;
+
+const images = {
```
BarnabyBishop, line 174: I should have removed this earlier when we added cloud images, but can you please remove this `images` from both the document types? 

Ideally we don't want any content images stored locally in the repo, we'd prefer them in cloud. Removing this kinda enforces it.

docs/src/components/document-renderer.tsx
```diff
@@ -97,13 +99,26 @@ const getRenderers = (
...
       );
     },
-    image: ({ src, alt }) => (
-      <img
-        className="rounded-lg my-2"
-        src={`/images/content/${slug}/${src}`}
-        alt={alt}
-      />
-    ),
+    image: ({ src, alt, title }) => {
```
BarnabyBishop, line 102: WRT my other comment of removing images from the document editor, perhaps we can add "Caption" to the cloud image?

BarnabyBishop (changes requested):
Looking shmick

docs/src/components/document-renderer.tsx
jtbrolin: Great point!

docs/keystatic.config.tsx
jtbrolin: Makes sense, cheers 🙌🏻

docs/src/content/pages/quick-start.mdoc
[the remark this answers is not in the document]
jtbrolin: Ah yeah that's not optimal 😅 But I agree with leaving it for consistency w peers for now 👍🏻

docs/src/content/authors/simon-swiss.yaml
```diff
@@ -0,0 +1,2 @@
+name: Simon Swiss
```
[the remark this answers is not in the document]
jtbrolin: Good point!

@simonswiss would you like to be rendered as a co-author for this first blog post and if so, would you like your full name?

docs/src/content/blog/the-fresh-keystatic-cli.mdoc
jtbrolin: Great point – thanks @simonswiss 🙌🏻

docs/src/components/navigation/header-nav.tsx
```diff
@@ -38,9 +47,30 @@ export function HeaderNav({
...
+              )}
+              href="/docs"
+            >
+              Docs
+            </Link>
+            <Link
+              className={cx(
+                linkStylesShared,
+                pathname?.startsWith('/blog')
+                  ? linkStylesCurrent
```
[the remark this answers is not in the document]
jtbrolin: **Great** catch! It took me a minute to understand what you meant, but I have totally missed that inconsistency, and I really like the styles/structure I have now for current page indication 🙌🏻

docs/src/app/(public)/blog/layout.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+export default async function RootLayout({
+  children,
+}: {
+  children: React.ReactNode;
+}) {
+  return (
+    <>
+      <main className="max-w-7xl mx-auto w-full px-6 pt-10 lg:pt-16 pb-16 flex flex-1">
+        <div className="mx-auto w-full sm:max-w-xl flex flex-col gap-4">
```
ChrisLaneAU, line 16: Sorry, I should have asked this before - what's the reason for having this `div` and not rolling the desired styles into the `main` above?

jtbrolin: Ahh good catch, it's not needed indeed!

docs/src/app/(public)/blog/page.tsx
```diff
@@ -0,0 +1,77 @@
...
+      return {
+        ...post,
+        parsedDate,
+        formattedDate,
+      };
+    })
+    .sort((a: { parsedDate: Date }, b: { parsedDate: Date }) =>
+      a.parsedDate < b.parsedDate ? 1 : -1
+    )
+    .filter((post: { entry: { draft: boolean } }) => !post.entry.draft);
```
ChrisLaneAU, line 29: If you move this `filter` up one level (i.e. chain it onto the `map`, then chain the `sort` to the `filter`) you will have less posts to `sort`.

Alternatively, you could switch the `map` for a `reduce` and handle the filtering during the same loop (see suggestion below).

I doubt we'll get to a point where we have so many blog posts that a more efficient sorting algorithm than `.sort()` will be needed but I think it'd be nice to improve the efficiency a little, with either just moving the `sort` or doing the below.

Let me know what you think of this (the filtering happens in the return ternary):

```suggestion
 type BlogPost = (typeof blogPosts)[number] & {
    formattedDate: string;
    parsedDate: Date;
  };

  const sortedPublishedBlogPosts = blogPosts
    .reduce<BlogPost[]>((acc, post): BlogPost[] | [] => {
      const { parsedDate, formattedDate } = parseAndFormatPublishedDate(
        post.entry.publishedOn
      );

      return post.entry.draft
        ? acc
        : [
            ...acc,
            {
              ...post,
              parsedDate,
              formattedDate,
            },
          ];
    }, [])
    .sort((a, b) => (a.parsedDate < b.parsedDate ? 1 : -1));
```

ChrisLaneAU (approved):
Nice! I'll have to bookmark this :)

docs/src/app/(public)/blog/page.tsx
BarnabyBishop: Good find to move the filter above the sort.

IMO `.map.filter.sort` is much more readable than a `reduce`. For reduce to be more optimised we likely have to have tens of thousands or hundreds of thousands of blog posts. 

Also in production this code is only run when the site is built, so I'd take readability over efficiency in this case.

docs/keystatic.config.tsx
```diff
@@ -82,6 +82,11 @@ export const componentBlocks = {
         label: 'Sizes',
         description: 'Optionally override the defualt sizes',
```
BarnabyBishop, line 83: This is not from your changes but I saw it so have to fix it 😂 

```suggestion
        description: 'Optionally override the default sizes',
```

docs/src/app/(public)/blog/page.tsx
ChrisLaneAU: No problem, sounds good!

jtbrolin: Ahh thanks @ChrisLaneAU! Doing the filtering before the sorting definitely make sense, thanks for catching that 🙏🏻 

I had to brush up a little about the `reduce()` method to understand your suggestion so thanks for writing that up!! Forever learnin' 🙌🏻  Chatted w Barney in standup as well and I think I'll stick with map + filter cause it feels easier to read, but defo valuable to know that the reduce would be more performant in this case

ChrisLaneAU: Out of curiosity about where the crossover point would be I ran some jsbench tests and `.map.filter` was always faster than `.reduce` by itself. I tested 3, 300, 3000, 30 000, 300 000.

The code comparison was (I adjusted the `length` for different size data):
```
Array.from({length: 100}).flatMap(emptyVal => ([{a:{b: false}}, {a:{b: true}}, {a:{b: false}}])).map(val => val).filter(val => val.a.b);

Array.from({length: 100}).flatMap(emptyVal => ([{a:{b: false}}, {a:{b: true}}, {a:{b: false}}])).reduce((acc, val) => val.a.b ? [...acc, val] : acc, []);
```

It's not exactly concrete research but it's interesting to see! So yeah, stick with `.map.filter.sort` here and take any advice I give with a pinch of salt 😅

jtbrolin: Oh okay very interesting indeed!!

BarnabyBishop: > Out of curiosity about where the crossover point would be I ran some jsbench tests and `.map.filter` was always faster than `.reduce` by itself. I tested 3, 300, 3000, 30 000, 300 000.

I love that you tested this rather than just accept my hand-wavy made-up stat 😂 

It is funny we can never tell how the JS engine is optimised under the hood for these situations. Good research!

BarnabyBishop (approved):

simonswiss (approved):

PR #385, opened 2023-06-29
Astro guide updates + content collections walkthrough

simonswiss:


simonswiss:
I'll go ahead and merge since this is definitely an improvement over what's currently live — feel free to suggest edits later, but I don't think we should shy away from sharing stuff early if it makes the overall documentation story better 👍

PR #389, opened 2023-06-29
Update ui.tsx to allow localhost has an origin

### LINKED ISSUE #366: Incorrect redirection to IP address 127.0.0.1 in Keystatic
When I access my project in dev mode at http://localhost:3000/keystatic, I always get redirected to http://127.0.0.1:3000/keystatic and receive an ERR_CONNECTION_REFUSED error from my browser. I think this is because I'm using Windows WSL2, and the IP address 127.0.0.1 doesn't work by default.

I tested this by adding it to an existing Next.js 13 project and a new template project from Keystatic CLI, and both encountered this error.

Why is Keystatic redirecting to this IP? Did I miss some configuration?

Config File (Same as in the docs):
```ts
// keystatic.config.ts
import { config, fields, singleton } from '@keystatic/core'

export default config({
    storage: {
        kind: 'local'
    },
    singletons: {
        homepage: singleton({
            label: 'Homepage',
            path: 'src/content/_homepage',
            schema: {
                headline: fields.text({ label: 'Headline' })
            }
        })
    }
})
```

fezproof:
This will allow Astro integration that doesn't need changing of the server host.

Closes #366

### DIFF
--- a/packages/keystatic/src/app/ui.tsx
+++ b/packages/keystatic/src/app/ui.tsx
@@ -267,17 +267,6 @@ export function Keystatic(props: {
   ) {
     throw new Error('Missing storage.repo.owner or storage.repo.name');
   }
-  useEffect(() => {
-    if (window.location.hostname === 'localhost') {
-      window.location.href = window.location.href.replace(
-        'localhost',
-        '127.0.0.1'
-      );
-    }
-  }, []);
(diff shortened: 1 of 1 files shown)


fezproof:
Actually closing this PR as I see the ip stuff is more engrained in the ecosystem than I thought. Possibly should include a way to configure it instead 🤔

PR #394, opened 2023-06-30
KEY-163: Improve keystatic.com metadata

jtbrolin:
- Adds dynamic metadata on /docs and /blog pages using the [generateMetadata function](https://nextjs.org/docs/app/api-reference/functions/generate-metadata#generatemetadata-function)
- Adds a `summary` field to /docs pages in /keystatic (like the /blog pages) so that a custom metadata description can be provided

NOTE: [Dynamically generated OG images](https://nextjs.org/docs/app/api-reference/functions/generate-image-metadata) will be added in future, but not included in this PR


docs/src/app/(public)/blog/[...slug]/page.tsx
```diff
@@ -87,3 +88,34 @@ export async function generateStaticParams() {
...
+  const slugPath = params.slug;
+  const slug = slugPath.join('/');
+
+  const page = await reader.collections.blog.read(slug);
+
+  const parentOGImages = (await parent).openGraph?.images || [];
+  const parentTwitterSite = (await parent).twitter?.site ?? '';
+
+  return {
+    title: page?.title,
```
Comment (ChrisLaneAU) [inline docs/src/app/(public)/blog/[...slug]/page.tsx:105]: Do we need fallbacks for any of these values if `page` is `undefined`? Or a different type via a `required` sort of thing in the schema?

docs/src/app/(public)/blog/[...slug]/page.tsx
```diff
@@ -87,3 +88,34 @@ export async function generateStaticParams() {
...
+    description: page?.summary,
+    openGraph: {
+      title: page?.title,
+      description: page?.summary,
+      url: `https://keystatic.com/blog/${slug}`,
+      type: 'article',
+      images: ['', ...parentOGImages],
```
Comment (ChrisLaneAU) [inline docs/src/app/(public)/blog/[...slug]/page.tsx:112]: What does the `''` do here?

docs/src/app/(public)/docs/layout.tsx
```diff
@@ -5,8 +5,10 @@ import { DocsFooter } from '../../../components/footer';
 import { getNavigationMap } from '../../../utils/reader';
 
 export const metadata = {
-  title: 'Keystatic - Docs',
-  description: 'Documentation for Keystatic.',
+  title: {
+    template: '%s - Docs | Keystatic',
+    default: 'Docs',
```
ChrisLaneAU, line 10: Not `Docs | Keystatic`?

ChrisLaneAU (commented):
This is really cool 😎

docs/src/app/(public)/blog/[...slug]/page.tsx
Comment (jtbrolin) [reply docs/src/app/(public)/blog/[...slug]/page.tsx:105]: Hmm I'm a bit unsure re: this.. I wonder if what they say on https://nextjs.org/docs/app/api-reference/functions/generate-metadata#default means that if `title: page?.title,` here is undefined then it will grab the `title.default` defined in the `/blog/layout.tsx`?

docs/src/app/(public)/blog/[...slug]/page.tsx
Comment (jtbrolin) [reply docs/src/app/(public)/blog/[...slug]/page.tsx:112]: Ah it does nothing I think, I left it there because I plan to add a custom OG image for each blog post and this is where I'd put that in. Is it better to remove that for now, and just do `images: parentOGImages,` you reckon?

docs/src/app/(public)/blog/[...slug]/page.tsx
Comment (ChrisLaneAU) [reply docs/src/app/(public)/blog/[...slug]/page.tsx:105]: It looks like if we do `title: page?.title` and that evaluates to `title: undefined` then the `<title>` is removed. But, if we conditionally add `title` then it does fall back as you're saying. I think we should change this to:

```
...(page?.title && { title: page.title })
```

What do you think?

docs/src/app/(public)/blog/[...slug]/page.tsx
Comment (ChrisLaneAU) [reply docs/src/app/(public)/blog/[...slug]/page.tsx:112]: Yeah, I think capture the intention in a ticket and we can come back to it. And make this `images: parentOGImages` as you suggest 👍

Comment (jtbrolin) [reply docs/src/app/(public)/blog/[...slug]/page.tsx:112]: That makes total sense, will do!

docs/src/app/(public)/blog/[...slug]/page.tsx
Comment (jtbrolin) [reply docs/src/app/(public)/blog/[...slug]/page.tsx:105]: Riight thanks for testing that!! Will change as per your suggestion 🙌🏻

Comment (ChrisLaneAU) [reply docs/src/app/(public)/blog/[...slug]/page.tsx:105]: No worries! We should examine the rest of the properties, too. And mimic the changes on the `docs` pages.

BarnabyBishop (commented):
Looking good, just the small things Laney suggested and I have one Q

docs/src/app/(public)/blog/[...slug]/page.tsx
```diff
@@ -87,3 +88,34 @@ export async function generateStaticParams() {
     slug: slug.split('/'),
   }));
 }
+
+export async function generateMetadata(
```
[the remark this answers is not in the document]
Comment (jtbrolin) [reply docs/src/app/(public)/blog/[...slug]/page.tsx:92]: Ahh yes can do! Good idea 🙌🏻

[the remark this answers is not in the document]
Comment (BarnabyBishop) [reply docs/src/app/(public)/blog/[...slug]/page.tsx:92]: As discussed typing this will be a bit of a pain to make it generic, for little pay-off. Let's leave it for now

docs/src/app/(public)/docs/layout.tsx
jtbrolin: Doesn't seem to be needed 🤔 If I do that then it renders a title with `Blog | Keystatic | Keystatic` if the `/blog/[...slug]/page.tsx` has no meta data `export const metadata = {};`

So I guess it takes the default, and then the template from the parent layout which is just `template: '%s - | Keystatic',`

docs/src/app/(public)/blog/[...slug]/page.tsx
Comment (jtbrolin) [reply docs/src/app/(public)/blog/[...slug]/page.tsx]: Don't know if this way is better but I've added fallbacks in a different way now:
```
const parentTitle = (await parent).title ?? 'Blog';
const title = page?.title ?? parentTitle;
```

Looks to work nicely when page?.title is undefined, but let me know whatya think!

simonswiss:
Just catching up on all this right now, looking good and good review suggestions 👍

BarnabyBishop (approved):
🚚

docs/src/app/(public)/docs/layout.tsx
ChrisLaneAU: >  it takes the default, and then the template from the parent layout

Ahh yes, you're absolutely right.

Re-reading the [Next docs on metadata](https://nextjs.org/docs/app/api-reference/functions/generate-metadata#template):

> `title.template` applies to child route segments and not the segment it's defined in.

docs/src/app/(public)/blog/[...slug]/page.tsx
Comment (ChrisLaneAU) [reply docs/src/app/(public)/blog/[...slug]/page.tsx]: Yeah, that's solid, nice work 🙌

ChrisLaneAU (approved):
🙌

PR #395, opened 2023-06-30
App header

jossmac:
Layout refactor that moves some stuff out of the sidebar into an app header.


packages/keystatic/src/app/shell/topbar.tsx
```diff
@@ -0,0 +1,555 @@
...
+        isHidden={{ below: 'tablet' }}
+      >
+        {config.storage.project}
+      </Text>
+      <Slash />
+      <BranchPicker />
+      <GitMenu />
+      <Box flex="1" />
+      <ThemeMenu />
+      {/* <UserMenu /> */}
```
jossmac, line 105: We don't have user data from cloud yet.

PR #396, opened 2023-06-30
KEY-165: Added filtering section to Reader API docs

jtbrolin:


docs/src/content/pages/reader-api.mdoc
```diff
@@ -52,13 +52,41 @@ const post = await reader.collections.post.read(slug);
...
 
-### For singletons
+### Filtering based on a field's value
+
+A common use case is to filter the entries based on a field's value before rendering them in the UI. You can use Keystatic's reader functions along with standard JS array methods like `.filter()` and `.sort()` to achieve this pattern.
+
+Let's say you're building a list page for blog posts, and the collection schema for each entry defines a `draft` checkbox field:
+
+```javascript
+// keystatic.config
```
ChrisLaneAU, line 68: We should add the file extension: `keystatic.config.js`

docs/src/content/pages/reader-api.mdoc
```diff
@@ -52,13 +52,41 @@ const post = await reader.collections.post.read(slug);
...
-const blog = await reader.collections.blog.all();
+const posts = await reader.collections.blog.all();
 ```
 
-### For singletons
+### Filtering based on a field's value
+
+A common use case is to filter the entries based on a field's value before rendering them in the UI. You can use Keystatic's reader functions along with standard JS array methods like `.filter()` and `.sort()` to achieve this pattern.
+
+Let's say you're building a list page for blog posts, and the collection schema for each entry defines a `draft` checkbox field:
```
ChrisLaneAU, line 65: Just to keep the scenario super obvious, and highlight _why_ you would want to filter the posts, it would be good to also say that the list page only shows published posts. Something like:

```md
Let's say you're building a list page for blog posts, and you only want to display published posts. The collection schema for each entry defines a `draft` checkbox field, indicating a not-to-be-published post:
```

The schema entry could tie in the scenario with repeated language (i.e. this would be line 73):

```js
draft: fields.checkbox({
  label: 'Convert to draft,
  description: 'Prevent publishing of this post. This can be changed later.',
  defaultValue: false,
})
```

Let me know what you think.

ChrisLaneAU (commented):
Couple of small suggestions but this is a really helpful addition to the docs, nice one 🙂

docs/src/content/pages/reader-api.mdoc
BarnabyBishop: This is a tricky one as there are many cases where it wouldn't be `.js`.

All of our starter templates and CLI use typescript, but even then it isn't as simple as using `.ts` as most of them actually use `.tsx`.

This is actually incorrect as _some_ have JSX in them for custom document fields, but most don't, so they probably should be `.ts` until it is needed to change it to `.tsx`. 

So really to be accurate with the file extension it should be `//keystatic.config.[js|jsx|ts|tsx]` 😱

BarnabyBishop: Thinking further we should probably decide on:

- Whether examples are in Typescript or Javascript and stick to it
- Decide how we reference files in code examples (like you flagged)

`.js` may be the right choice, but feels a little weird as all our ways to get started it in TS.

ChrisLaneAU: > Thinking further we should probably decide on:

> Whether examples are in Typescript or Javascript and stick to it
> Decide how we reference files in code examples (like you flagged)

Yes 💯 ! Ultimately it needs to be JavaScript so that seems like the best baseline and we shouldn't be presumptuous about folks' superset/language. It would be nice to offer [JS and TS examples, similar to how Next JS does in their docs](https://nextjs.org/docs/app/api-reference/functions/revalidatePath).

ATM, though, the most responsible thing to do may be sticking to TS as there's probably better value we can offer the community by concentrating on other things while still being available to help anyone convert TS examples to JS. And later on we come back to expand the examples with JS, and offer a JS option in the CLI.

Good call out, I hadn't fully considered this 🙌

jtbrolin: Great thoughts and agreed – need to be consistent and would be v nice to add option for code examples in both js and ts in the future 😍

docs/src/content/pages/reader-api.mdoc
jtbrolin: Great suggestions!! Thanks Laney, will update

BarnabyBishop (approved):
🚢

PR #413, opened 2023-07-05
KEY-163: Adds summaries to docs pages

jtbrolin:
- Adds summaries to docs pages, used for the metadata descriptions


docs/src/content/pages/collections-and-singletons.mdoc
```diff
@@ -1,5 +1,8 @@
 ---
 title: Collections & singletons
+summary: >-
+  Keystatic has two concepts or mechanisms to manage data: collections and
+  singletons.
```
BarnabyBishop, line 5: A bunch of these summaries have an extra line break for the last word or last few words. Is that intentional?

jtbrolin: Nah let me go through and see if I can fix that 🙏🏻

jtbrolin: Actually, I added all these through the Admin UI so it seems if the string is a certain length then it will populate the mdoc like this. Still think we need to tweak these @BarnabyBishop?

BarnabyBishop: Probably not worth it if the AdminUI will just wrap them again when we make changes. Good to leave for now

BarnabyBishop (approved):

PR #414, opened 2023-07-05
KEY-138: TOC's active style

jtbrolin:
- My latest attempt that I described in standup: Intersection observer looks at currently visible headings in the viewport and sets the first one as the "active-styled" one
- Adds a persistent #overview heading for the H1 on each docs' page
- Hard-code fixes a few duplicate headings, as we don't currently have logic to prevent identical heading id's


jtbrolin:
Here's the other branch I was playing with 2 intersection observers but I didn't like how it was turning out so abandoned @ChrisLaneAU @BarnabyBishop https://github.com/Thinkmill/keystatic/blob/toc-active-style/docs/src/components/navigation/table-of-contents.tsx

PR #419, opened 2023-07-05
Fix typing in installation-astro.mdoc

fluriumteam:
src/pages/eystatic/[...params].astro -> src/pages/keystatic/[...params].astro


fluriumteam:
Pull request already exists

PR #436, opened 2023-07-09
Organise examples

JedWatson:
Some repo cleanup around examples and the main dev project

* make the `keystatic` folder in the repo just another app, and calls them `examples` because I think that's closer to what they are
* adds `examples` to codeowners
* rename `local-config.ts` to `keystatic.config.ts` in line with the rest of the convention
* add local development instructions to the main readme

note [the keystatic app on vercel](https://vercel.com/thinkmill-labs/keystatic) will need its settings changed when this is merged, and preview deployments are broken in the meantime because the start folder has moved


examples/localization/src/config.ts
```diff
@@ -1,5 +1,5 @@
 import { collection, config, fields } from '@keystatic/core';
-import { locales } from '../../../packages/keystatic/src/app/l10n/locales';
+import { locales } from '@keystatic/core/src/app/l10n/locales';
```
emmatown, line 2: This weird import was intentional because `@keystatic/core/src/app/l10n/locales` won't work because of the `exports` field

PR #437, opened 2023-07-10
Resizable panels

jossmac:
Allow users to resize parts of the UI:
- the sidebar navigation pane
- secondary fields, when `entryLayout='content'`


emmatown (approved):
yeah... we really need to refactor some things but eh 🤷‍♀️, all this is probably fine

PR #458, opened 2023-07-17
KEY-196: Proposed revamp of docs navigation sections

simonswiss:
This PR re-organises some of the docs pages and phases out the "Learn Keystatic" section, in favour of two new sections: "Core concepts" and "Reference".

I think this makes for a little more sense based on the current content on those pages.

There will always be room for "Learn Keystatic", but this may be in the form of a "Resources" section that links to some tutorials, screencasts etc. I don't think the "Learn Keystatic" made sense as the content evolved in those docs.

I also broke down "Collections & Singletons" into two separate pages. And removed the "FAQ" page which had mostly only out-of-date content.

Totally open for discussion, this is just a proposal!

CleanShot 2023-07-17 at 15 35 35@2x


JedWatson (approved):
Nice, I like it 👍

PR #465, opened 2023-07-18
Hide dashboard header on desktop

JedWatson:
@jossmac this is a pretty blunt prop to add to the `AppShellHeader` component...

... but the dashboard header has been annoying me on desktop views, it feels like useless decoration (compared to the list and entry views where the toolbar has useful navigation / actions in it)

So this approach hides it on above the mobile breakpoint, but shows it when the navigation collapses, which means we keep the header and UI to show the sidebar menu on mobile.

image

### DIFF
--- a/.changeset/polite-lizards-share.md
+++ b/.changeset/polite-lizards-share.md
@@ -1,0 +1,5 @@
+---
+"@keystatic/core": patch
+---
+
+Hide dashboard header on desktop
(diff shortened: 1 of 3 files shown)


jossmac (changes requested):
I don't think this is a good idea: when the user collapses the sidebar on the dashboard page they won't be able to get it back.

PR #466, opened 2023-07-18
Update panel size max for sidebar

JedWatson:
This isn't a particularly big change or comprehensive fix or anything, but it's been annoying me that when I resize windows, the navigation sidebar always ends up too big and then I have to drag it smaller.

So after some tinkering (you're right @jossmac this is really fiddly) I ended up just reducing the maximums to make it "feel better". Let's shrink it and see if anyone comes up with a use-case for it to be meaningfully wider than the new max?

Otherwise it really eats into content entry space when you resize the window, on a normal 13" screen viewport.

**Before**:

image

**After**:

image

### DIFF
--- a/.changeset/few-rockets-invite.md
+++ b/.changeset/few-rockets-invite.md
@@ -1,0 +1,5 @@
+---
+"@keystatic/core": patch
+---
+
+Reduced sidebar panel max size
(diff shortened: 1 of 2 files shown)


jossmac (approved):
Yeah, the percentage API isn't appropriate. Unfortunately the author will not support pixel dimensions: [46](https://github.com/bvaughn/react-resizable-panels/issues/46), [47](https://github.com/bvaughn/react-resizable-panels/issues/47), [78](https://github.com/bvaughn/react-resizable-panels/issues/78), [114](https://github.com/bvaughn/react-resizable-panels/issues/114)...

I need to roll a bespoke implementation, but it's not priority.

PR #477, opened 2023-07-24
Split view

jossmac:
Export `SplitView` components from new "split-view" package.

> A split view is two sections or "panes", including a moveable separator between them that allows users to change the relative size of the panes.


design-system/pkg/src/split-view/SplitView.tsx
```diff
@@ -0,0 +1,504 @@
...
+  });
+
+  const wrapperRef = useRef<HTMLDivElement>(null);
+  const offsetRef = useRef(0);
+  const moveRef = useRef(0);
+
+  // reverse drag logic when the primary pane is on the right or if the locale
+  // direction is right-to-left
+  const isReversed = useMemo(() => {
+    // @ts-expect-error FIXME: this feels super dodgy
```
emmatown, line 93: tbh, personally I would forgo trying to do the compositional api thing here of seperate components and just make it one thing (not a blocking thing, just a thought if it's helpful)

emmatown (approved):
really cool!!

design-system/pkg/src/split-view/SplitView.tsx
jossmac: resolved by ec140e08a0b67b9dbad8eaceaf5ba97f6ac8f3f0

PR #479, opened 2023-07-24
Dial down expectations of things working before adding API routes

simonswiss:
CleanShot 2023-07-25 at 09 59 38@2x


emmatown (approved):
(personally I wouldn't have a screenshot of it seemingly working before it actually works but sure)

simonswiss:
This 100% makes sense — that screenshot here is more confusing than anything. Removed, and it reads much better. Thank you!

docs/src/content/pages/installation-next-js.mdoc
```diff
@@ -102,17 +102,9 @@ export default function Page() {
 
 ```
 
-Great! So now, we should be able to visit the `/keystatic` router in the browser.
+We're not done yet, but we should be able to visit the `/keystatic` route in the browser.
```
emmatown, line 105:
```suggestion
```

(I wouldn't invite them to try using it if they need to do more things before it actually works but again, don't mind strongly)

emmatown (approved):

PR #498, opened 2023-07-31
Update Astro's Markdoc integration and Astro (bug in previous version)

simonswiss:
Not sure if this warrants a changeset — probably a `patch`?


emmatown (commented):
To be released, yes this needs a changeset, a patch makes sense

emmatown (approved):

PR #509, opened 2023-08-08
Added missed folder 

teinett:
src/content/posts/*.mdoc


JedWatson:
Good catch, thanks @teinett!

simonswiss:
Thank you for this!

PR #511, opened 2023-08-08
Update image.mdoc for a spelling mistake

NikharPandya:
"in" instead of "it"


simonswiss:
Thanks for this, @NikharPandya 🙏

PR #514, opened 2023-08-10
Disallow whitespace in slugs

### LINKED ISSUE #495: Trim space at the end of url
Content team can make a mistake of having space at the end of the url slug and it breaks the page. (404 throws up)

PS: @matojdowski faced this issue while working.

JedWatson:
Fixes #495

I went back and forth on how to solve the problem described in #495, ended up deciding for now to keep it simple and add always-on validation for slug fields that ensures they don't contain whitespace.

Generally, whitespace might create problems between URLs and different filesystems, and I can't think of a use case where you'd want it (and in that case, could always use a text field anyway) so it seems pretty safe given the use-case to just prevent it entirely.

We might want to either be stricter (preventing any characters that get escaped in URLs) or less strict (allow customisation of the validation rules) but I think let's start here and fix the footgun and see what feedback we get.


emmatown (commented):
Could we start with only disallowing leading and trailing whitespace?

emmatown:
If I had some arbitrary files that I wanted to manage with Keystatic and they already had spaces in them, this would just block me from doing that in Keystatic (though annoyingly only _sort of_, because you'd be able to see them and open the item and make changes but you'd be blocked from saving)

emmatown (approved):

PR #516, opened 2023-08-10
Add --host flag to astro template

simonswiss:


Maarten-Dekker:
Only appending --host doesn't work for me (windows no wsl). I think a more explicit approach would fix this forever:
astro dev --host 127.0.0.1

simonswiss:
> Only appending --host doesn't work for me (windows no wsl), as this will provide the following IP's:
> 
> ```
> Local    http://localhost:3000/
> Network  http://192.168.1.182:3000/
> ```
> 
> I think a more explicit approach would fix this for everyone: `astro dev --host 127.0.0.1`

I've updated the script — thank you!

JedWatson:
CI issues on this branch were due to the lockfile changes, so I've reverted them and now the lint/type checks pass as expected on `main`.

It looks like they're not strictly necessary to this change, although I've also noticed the `astro` template is out of date by a minor version (`2.9.6`, latest is `2.10.7`)

So I suggest we separate this change from updating the lock file, which we can do in a separate PR, and also bump the astro version separately because there's some other work going on to update the astro integration.

JedWatson (approved):

PR #522, opened 2023-08-13
Update card hover style

JedWatson:
@jossmac continuing the riff on styling, agree with you the `accent` colour for borders on hover isn't quite where it's at, and I couldn't find anything else I prefer aesthetically than this.

If you're also "this is better", let's go with the `emphasis` token for now, and pick it up back in design with @tuan23

### DIFF
--- a/packages/keystatic/src/app/dashboard/components.tsx
+++ b/packages/keystatic/src/app/dashboard/components.tsx
@@ -78,7 +78,7 @@ export const DashboardCard = (
                 textDecorationColor: tokenSchema.color.foreground.neutral,
 
                 '::before': {
-                  borderColor: tokenSchema.color.border.accent,
+                  borderColor: tokenSchema.color.border.emphasis,
                 },
               },
               '&:active': {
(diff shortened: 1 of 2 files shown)


packages/keystatic/src/app/dashboard/components.tsx
```diff
@@ -78,7 +78,7 @@ export const DashboardCard = (
                 textDecorationColor: tokenSchema.color.foreground.neutral,
 
                 '::before': {
-                  borderColor: tokenSchema.color.border.accent,
+                  borderColor: tokenSchema.color.border.emphasis,
```
jossmac, line 81: You want to keep `accentEmphasis` when [pressed](https://github.com/Thinkmill/keystatic/pull/522/files#diff-21caac9c730e2c2bbc5aad689480b464e974d474b2bf7f0b702d87f6a748dae7R86)?

JedWatson: Yep, lgtm

Issue #531, opened 2023-08-15
Better DX / documentation for the Document field


JedWatson:
Our Document field is a bit obscure to get running successfully with, a few things we can do to improve it:

- [x] Only include valid markdown formatting options when you're using [shorthand syntax](https://keystatic.com/docs/fields/document#usage-example)
- [x] Document what the shorthand syntax expands into, and include an example of the **full** [reference of fields](https://docsmill.dev/npm/@keystatic/core@0.0.115#/.fields.document.FormattingConfig) that are available
- [ ] Explain that turning on other features will require custom tags in your MarkDoc rendering, ref #504
- [ ] Add dev example projects that show how to render the full set of formatting options, in Next.js (with Reader) and Astro (with Markdoc integration)

emmatown:
Closing since the document field is now deprecated and we have better docs for the new editor fields

PR #538, opened 2023-08-15
Adding changeset for document field changes

### LINKED ISSUE #504: Underline option throws Astro error
When using the option to underline text inside the document editor, Keystatic produces this markup inside the mdoc file: `{% u %}lorem ipsum{% /u %}`

I use the `format: { contentField: 'content' }` option.

Trying to render this via Astro's content collection throws an error:

```
 error   **src/content/kontaktCTA/index.mdoc** contains invalid content:
  - Undefined tag: 'u'
```
I am pretty sure there is no underline feature in markdoc, so disabling this by default  in the editor might be a good idea.

JedWatson:
These changes were merged to `main` without a changeset, so catching up with this PR (and will review branch protection rules tomorrow!)

**TL;DR** I've updated the document field defaults, and the associated documentation, as mentioned in #531

This closes #504

I'd like to spend some more time with the docs to better explain how to set up custom tags in Markdoc properly if you opt into the (now-off) additional formatting options, so #531 remains open until then, but this means we have one less footman in the meantime.

### DIFF
--- a/.changeset/three-bears-weep.md
+++ b/.changeset/three-bears-weep.md
@@ -1,0 +1,41 @@
+---
+'@keystatic/core': patch
+---
+
+The `document` field defaults for formatting have changed to exclude options that require custom Markdoc tags.
+
+See https://keystatic.com/docs/fields/document#formatting-options for the new defaults.
+
+When updating, if you have configured a document field with shorthand for the `formatting` config:
+
+```ts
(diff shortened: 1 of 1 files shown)


emmatown (approved):
It'll be kinda breaking in probably subtly annoying ways but I broadly agree

Issue #529, opened 2023-08-15
Document `path` glob syntax


jossmac:
Related to #519

Document `path` glob syntax, probably here: https://keystatic.com/docs/how-keystatic-organises-your-content

```js
export default config({
  collections: {
    posts: collection({
      label: 'Posts',
      path: 'content/posts/**/*',
      // ...
    })
  }
})
```

PR #539, opened 2023-08-16
Content path and * wildcard documentation

simonswiss:
Improves and extends the documentation of the `path` option on singletons and collections.

Fixes https://github.com/Thinkmill/keystatic/issues/529

### DIFF
--- a/docs/src/content/pages/how-keystatic-organises-your-content.mdoc
+++ b/docs/src/content/pages/how-keystatic-organises-your-content.mdoc
@@ -74,7 +74,7 @@ path: 'packages/design-system/*/docs/'
 ### Nested slug example
 
 {% aside icon="👉" %}
-  `path: 'content/posts/**/*'`
+  `path: 'content/posts/**'`
 {% /aside %}
 
 There may be situations where you need the `slug` of an entry to be following a multi-folder structure.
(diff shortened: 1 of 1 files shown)


docs/src/content/pages/how-keystatic-organises-your-content.mdoc
```diff
@@ -35,33 +35,84 @@ export default config({
...
+  ├── en  
+      └── post-1.mdoc
+  ├── fr  
+      └── post-1.mdoc
+```
+
+You can enable this by using the `**/*` wildcard in your `path`: 
+
+```
+path: 'content/posts/**/*'
```
emmatown, line 96:
````suggestion
You can enable this by using the `**` wildcard in your `path`: 

```
path: 'content/posts/**'
````

simonswiss: Updated these 2 + another instance of `**/*` 👍

emmatown (approved):
This kinda confused me when I was trying to review it because I did not expect it to be under "Organised content" but having this documented somewhere seems better than the status quo so 🚢

JedWatson (approved):

JedWatson (approved):
Approved. Agree this would make more sense somewhere else, but we can change that in a future PR, this is great new content! thanks @simonswiss

simonswiss:
Thank you!

And yep, I agree this probably needs a dedicated page or API reference thing, rather than a core "concept".

PR #540, opened 2023-08-16
Proposal to move path configuration to dedicated page

simonswiss:


docs/src/content/pages/how-keystatic-organises-your-content.mdoc
```diff
@@ -8,95 +8,11 @@ Those are defined in the&nbsp;[Keystatic configuration](/docs/configuration).
...
-```
-
-You can enable this by using the `**` wildcard in your `path`: 
-
-```
-path: 'content/posts/**'
-```
-
-Since Keystatic `slugs` can contain `/` characters, you construct a multi-folder structure with `slug` field values like: `en/post-1` and `fr/post-1`.
+We'll go over core concepts here, but check out the [Path configuration page](/docs/path-configuration) for more details and examples.
```
simonswiss, line 15:
```suggestion
We will go over core concepts here, but check out the [Path configuration page](/docs/path-configuration) for more details and examples.
```

PR #545, opened 2023-08-17
docs: add section about disabling indexing

florian-lefebvre:
Following https://github.com/Thinkmill/keystatic/discussions/486

cc @simonswiss

### DIFF
--- a/docs/src/content/pages/installation-astro.mdoc
+++ b/docs/src/content/pages/installation-astro.mdoc
@@ -193,7 +193,67 @@ Notice we also opt out of prerendering for those routes.
 
 We should be all set to use Keystatic now.
 
-### Creating a new post
+## Disabling indexing
+
+When using the `local` strategy, chances are you won't want to include these routes in the production build.
+
+### Adding redirects
+
+This can be achieved by levering `Astro.redirect`. Here is how to do it:
+
+```diff
+---
+// src/pages/keystatic/[...params].astro
+import { Keystatic } from '../../../keystatic.page'
+
+export const prerender = false
+
++ if (import.meta.env.MODE === 'production') {
++   return Astro.redirect('/')
++ }
+---
+
+<Keystatic client:only />
+
+```
+
+```jsx
+// src/pages/api/keystatic/[...params].ts
+import { makeHandler } from '@keystatic/astro/api'
+import keystaticConfig from '../../../../keystatic.config'
+
+export const all = makeHandler({
+  config: keystaticConfig,
+})
+
+export const prerender = false
+
++ if (import.meta.env.MODE === 'production') {
++   return Astro.redirect('/')
++ }
+```
+
+### Excluding routes from sitemap
+
+If you're using `@astrojs/sitemap`, you can exclude those routes as well:
+
+```diff
+// astro.config.mjs
+import { defineConfig } from 'astro/config'
+import sitemap from '@astrojs/sitemap';
+
+export default defineConfig({
+  integrations: [
++     sitemap({
++       filter: (page) => !page.includes("keystatic"),
++     });
+  ]
+})
+```
+
+---
+
+## Creating a new post
 
 Try and visit the `/keystatic` page in the browser one more time, and click on the “Posts” collection.
 


docs/src/content/pages/installation-astro.mdoc
```diff
@@ -193,7 +193,67 @@ Notice we also opt out of prerendering for those routes.
...
 We should be all set to use Keystatic now.
 
-### Creating a new post
+## Disabling indexing
+
+When using the `local` strategy, chances are you won't want to include these routes in the production build.
+
+### Adding redirects
+
+This can be achieved by levering `Astro.redirect`. Here is how to do it:
```
simonswiss, line 202:
```suggestion
You can redirect visits to the `/keystatic` route in production by leveraging `Astro.redirect`:
```

docs/src/content/pages/installation-astro.mdoc
```diff
@@ -193,7 +193,67 @@ Notice we also opt out of prerendering for those routes.
...
+
++ if (import.meta.env.MODE === 'production') {
++   return Astro.redirect('/')
++ }
+---
+
+<Keystatic client:only />
+
+```
+
```
simonswiss, line 219:
```suggestion

You'll need to do the same for the `api/keystatic` routes:

```

docs/src/content/pages/installation-astro.mdoc
```diff
@@ -193,7 +193,67 @@ Notice we also opt out of prerendering for those routes.
 
 We should be all set to use Keystatic now.
 
-### Creating a new post
+## Disabling indexing
```
simonswiss, line 196:
```suggestion
## Preventing indexing of Keystatic routes in production
```

docs/src/content/pages/installation-astro.mdoc
JedWatson: Maybe some simpler language even:

```suggestion
You can redirect visits to the `/keystatic` route in production with `Astro.redirect`:
```

docs/src/content/pages/installation-astro.mdoc
JedWatson: I think it's worth being more specific here to avoid confusion:

```suggestion
## Preventing indexing of Keystatic Admin UI routes in production
```

simonswiss:
Hey @florian-lefebvre ✨ 

I know I asked you to open this PR in the Astro guide, but I sort of changed my mind on where this information should belong.

Last week and the conversations with the Astro docs team led us to a huge re-think of how we should approach guides, particularly framework integration guides. 

After a major rewrite of this guide with a sharp focus on avoiding any "side quests", I can't help but feel that this section about shutting down `/keystatic` routes in production should belong somewhere else. It just snaps the user out of the procedural, step by step installation.

I see this info in something like a recipe or a tutorial, for a specific need. It's great content and very useful when you need it, but IMO, it does not belong in the "get setup with Astro" guide itself.

I like what the [Astro docs](https://docs.astro.build/en/recipes/) do with recipes:

CleanShot 2023-08-23 at 21 37 06@2x

If you don't mind, I'll try to stand a "Recipes" section in the nav tomorrow and make your contribution one of these guides.

I think it will work better that way.

Sorry about the flip-flopping around 😅 I've become very passionate about keeping integration guides as short and to-the-point as possible, and move all tangential content to another place.

Let me know what you think!

florian-lefebvre:
Hey! No worries, I also think it makes more sense to get it out of the integration guide. I think recipes are more relevant so that it's  not mixed with "official" posts from the team.

Regarding the author link, one idea could be to add contributors to recipes like on nuxt v2 section (scroll all the way down on [this page](https://v2.nuxt.com/docs/get-started/upgrading)).

Wdyt?

simonswiss:
Gotta go to bed now, but I've pushed my code in this branch (did not have access to push in your fork): https://github.com/Thinkmill/keystatic/pull/579

That PR contains all the commit history of this PR here.

I also think the recipes is a better place. Right now all docs pages are the same content type in Keystatic, so adding an author would mean every docs page has an author field.

I think we can add a `contributed by` aside callout at the top of the recipe for that effect 👍

Feel free to make suggestions, and let's try to get this merged tomorrow!

PR #546, opened 2023-08-18
Document editor main content

jossmac:
Optimise the editor appearance when `entryLayout="content"` for a more focused experience.

Component library:

- Update the antialiasing behaviour everywhere 
- New `Prose` component from "@keystar/ui/typography" package.
- Improve `Field` implementation and types

### DIFF
--- a/packages/keystatic/src/form/fields/document/DocumentEditor/render-element.tsx
+++ b/packages/keystatic/src/form/fields/document/DocumentEditor/render-element.tsx
@@ -1,6 +1,5 @@
 import { RenderElementProps } from 'slate-react';
 
-import { BlockquoteElement } from './blockquote/blockquote-ui';
 import {
   ComponentBlocksElement,
   ComponentInlineProp,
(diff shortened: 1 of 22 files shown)


packages/keystatic/src/form/fields/document/DocumentEditor/render-element.tsx
```diff
@@ -143,11 +133,7 @@ export const renderElement = (props: RenderElementProps) => {
     default:
       let { textAlign } = props.element;
       return (
-        <p
-          className={blockElementSpacing}
-          style={{ textAlign }}
-          {...props.attributes}
-        >
+        <p style={{ textAlign }} {...props.attributes}>
```
jossmac, line 136: We should get rid of text-alignment as a feature.

JedWatson: This is now off by default as of #538

You can still turn it on though, and it adds a Markdoc attribute to the block

This seems fine, because aligning text is still a valid feature for content editing, but it's now an explicit opt-in (otherwise people would have to use custom component blocks to make it possible and that would be generally worse in all ways)

JedWatson (approved):

PR #551, opened 2023-08-18
Big Docs Revisit!

simonswiss:
~~Putting this in Draft because I ran out of time and kids are calling. Will pick up on Monday. Needs review, but~~ I think it *definitely* improves a lot of things.

The navigation on the left flows much more naturally, there are less confused and slightly overlapping, double-up pages.

~~A bit in WIP mode as I pushed when I ran out of time.~~

A pretty good Friday afternoon session, I gotta say!

## Rough list of what this PR does:

- Removed the `Local vs GitHub` page, broken down into `Local mode` and `GitHub mode` pages. The `GitHub mode` page is the former `Connect to GitHub` guide which wasn't placed ideally in the flow of the docs.

- Organised the `Organised content` and `path` mess into two pages: `Organised content` and `Path Wildcard`.

- Rewrote a ton of copy trying to make it simpler, less story-telly, more like a set of instructions.

- Created permanent redirects in the next config file for the old pages.

## Preview URL

https://keystatic-site-git-docs-iteration-thinkmill-labs.vercel.app/docs/introduction


simonswiss:
Reviewed internal links for all pages to make sure they are updated to the new pages — all is well!

simonswiss:
I am confident this radically improves the flow of the documentation, but since there are some "sweeping" changes, I'd like to at least bring some attention to it before merging.

That said — it would also be good to merge soon before this becomes a "monster PR" — and the other work I want to do (use Cloud images, work on copy on particular pages etc) would all benefit from this being merged first.

So @JedWatson I'll leave this open for a bit so you can have a look and pull the handbrake if you feel there's a need to, otherwise I am going to make a judgement call and merge it :)

Issue #444, opened 2023-07-12
New field: datetime


florian-lefebvre:
Currently, the date field only supports the following format: "yyyy-mm-dd". Would be great to have a Datetime field to support iso strings like `2021-03-03T19:44:47`

jossmac:
BTW @JedWatson We're most of the way there with the `DateField` https://keystar-ui.vercel.app/package/date-time/date-field. I just need to set aside some time to implement its companion `DatePicker` component that also includes a calendar dialog.

emilienbidet:
Hello @florian-lefebvre and @JedWatson ! I need this feature too. I would like to contribute to this project to make this available asap.

If I understood well the fields ui are generated using the keystar-ui package is it right ?

Could someone guide me on where can I found where to start in this codebase ?

Thanks 
Emilien

florian-lefebvre:
I think those are all the fields https://github.com/Thinkmill/keystatic/tree/main/packages/keystatic/src/form/fields

emmatown:
Added in #552

PR #552, opened 2023-08-18
Add datetime field

emilienbidet:
Hello Thinkmill and keystatic team 👋🏻

CleanShot 2023-08-18 at 13 41 27@2x

Following #444

I need the datetime field for my personal project so I just did it.
Thank you open source, this is my first real contribution to keystatic, so glad to work on this project.

### Changes Made

1. Created a new field named `datetime` to handle DateTime input using the `<input type="datetime-local">` form field.
2. Implemented the `DatetimeFieldInput` component for capturing Datetime input.
3. Added the `validateDateTime` function to handle DateTime validation.
4. Created a new demo component named `DatetimeFieldDemo` to showcase the usage of the `datetime` field.
5. Updated the documentation to include details about the `datetime` field, its usage, and type signature.

Please review the changes made and let me know if any further adjustments are needed.


JedWatson:
Hey @emilienbidet!

Thanks for the PR. It's late Friday evening here in Australia so we probably won't land this before early next week, but at a glance it looks good. Will get you a proper review soon!

emilienbidet:
Hi @JedWatson !

Thank you for your answer! Okay, have a nice weekend then!
I wait for your review.

FYI : I didn't handle **timezone**, I went straight to basic.

Thank you very much everyone,
Have fun

emilienbidet:
I'm wondering... Why for keystatic date field we don't use the `DateField` from the design-system?

Another implementation of this datetime field would be to just to update the date field to make it use the DateField from the design-system and add the granularity props as a field customisation properties like this :

```ts
datetime: fields.date({
  label: 'Date and time',
  description: 'The date and time of the event',
  granularity: 'day' | 'hour' | 'minute' | 'second',
  validation: {
    isRequired: true,
    min: '2023-08-15T00:00Z',
    max: '2023-08-20T00:00Z',
  },
}),
```

I prefer this approach because it match to the design-system and let the dev choose what type of date he wants.

JedWatson:
Hey @emilienbidet!

We took a look at this, and I also had a chat with @jossmac about adding Time support to the `DateField` in the design system.

We'd like to expand on the design system component so it can handle selecting times as well, but it's not going to happen quickly, and it should be fine to update the UI for this field in the future when the design system component is ready; the native type is less consistent but will get things unblocked and it wouldn't be a breaking change to switch them out.

In terms of having one field or two, we're inclined to go with two, because we've found it easier to manage each field type having a distinct data type that it serialises to; e.g we're planning to improve on slugs in the near future and one of the things that we want is to be able to encode date / datetime field values into them to make filesystem sorting easier. In that case, you'd handle the value differently, and having distinct field types makes that easier than one field that has a variable type.

So we're happy to go with this implementation for now!

JedWatson (approved):
Great PR, thanks @emilienbidet!

Very happy to have you contributing to Keystatic ❤️

PR #553, opened 2023-08-20
Add meta tag to prevent automatic zoom on mobile

JedWatson:
I've been doing some testing on mobile, and on iOS our Admin UI zooms in on the focused text field slightly, which makes for some weird layout effects in our responsive UI.

Adding this meta tag to the provider (at least in my testing on iOS 16) fixes the issue and makes the mobile experience feel more robust, although it doesn't actually prevent manual zooming (iOS ignores the instruction to prevent zooming entirely, but it does fix the auto-zoom on focus behaviour)

Preventing zooming entirely [isn't recommended](https://www.boia.org/blog/web-accessibility-tips-dont-disable-zooming-yes-even-on-mobile) for accessibility reasons, but this seems like an OK trade-off, since it doesn't _actually_ prevent zoom, and the user experience feels much less broken given the responsive UI otherwise works quite well.

Arguably, we should look at upping the input font size to 16px which would also prevent the default zooming behaviour cc/ @tuan23 @jossmac

### DIFF
--- a/packages/keystatic/src/app/provider.tsx
+++ b/packages/keystatic/src/app/provider.tsx
@@ -240,10 +240,7 @@ export default function Provider({
           href="https://fonts.googleapis.com/css2?family=Inter:wght@400;500;600;700&display=swap"
           rel="stylesheet"
         />
-        <meta
-          name="viewport"
-          content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no"
-        />
+        <meta name="viewport" content="width=device-width, initial-scale=1" />
         <UrqlProvider value={useMemo(() => createUrqlClient(config), [config])}>
           {children}
         </UrqlProvider>
--- a/packages/keystatic/src/app/provider.tsx
+++ b/packages/keystatic/src/app/provider.tsx
@@ -240,6 +240,10 @@ export default function Provider({
           href="https://fonts.googleapis.com/css2?family=Inter:wght@400;500;600;700&display=swap"
           rel="stylesheet"
         />
+        <meta
+          name="viewport"
+          content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no"
+        />
         <UrqlProvider value={useMemo(() => createUrqlClient(config), [config])}>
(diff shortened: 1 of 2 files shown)


packages/keystatic/src/app/provider.tsx
```diff
@@ -240,6 +240,10 @@ export default function Provider({
           href="https://fonts.googleapis.com/css2?family=Inter:wght@400;500;600;700&display=swap"
           rel="stylesheet"
         />
+        <meta
+          name="viewport"
+          content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no"
+        />
```
jossmac, line 246: Prefer:

```suggestion
        <meta name="viewport" content="width=device-width, initial-scale=1" />
```

jossmac (approved):

JedWatson:
@jossmac bad news: it doesn't solve the problem on mobile with your suggested change 🙁 

more bad news: this actually doesn't work anyway, at least in the docs site, where Keystatic is implemented as Next.js pages in the `app` folder. I did some digging on why my testing of this was turning up inconsistent results, and it turns out that Next.js automatically includes metadata fields with [conflicting defaults](https://nextjs.org/docs/app/building-your-application/optimizing/metadata#default-fields) so their default overrides the meta tag I'm adding here.

Adding this to the keystatic route in the docs page `docs/src/app/keystatic/layout.tsx`:

```
export const metadata: Metadata = {
  viewport:
    'width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no',
};
```

... actually (and properly, not inconsistently) fixes the problem on mobile. But because of how our integration works, this code exists in userland of every Keystatic implementation out there... and I'm not sure how to fix it within the code we control.

So this is going to take a bit more thought.

jossmac:
> bad news: it doesn't solve the problem on mobile with your suggested change 🙁

Okay. Please don't include `maximum-scale=1, user-scalable=no` anywhere.

ArnavK-09 (approved):

PR #554, opened 2023-08-20
UX improvements to the Cloud Image ComponentBlock

JedWatson:
This improves the workflow and UX for the `cloudImage` component block, but needs some design love before it's ready to merge.

When you first insert the component block, it now prompts you to focus the field and paste image data (can be a URL, JSON, or Markdown, it just has to have a valid image src in it)

image

image

Once you paste an image in, it will validate the image and detect the width and height, then switch to the current UI:

image

Depending on browser capabilities, sometimes we can trigger a paste event on click, otherwise there's an invisible text field that receives focus as a fallback. Might need some tweaking as we try different browsers... I've tested it on iOS, but need someone to check that the UX flow is good on Android as well. Firefox and Safari on MacOS are good.


packages/keystatic/src/component-blocks/cloud-image-preview.tsx
```diff
@@ -10,65 +10,189 @@ import { Icon } from '@keystar/ui/icon';
...
-export function CloudImagePreview(
-  props: PreviewProps<
-    ObjectField<typeof import('./cloud-image-schema').cloudImageSchema>
-  > & {
-    onRemove(): void;
-  }
-) {
-  const src = props.fields.src.value;
-  const alt = props.fields.alt.value;
+import { detect } from 'detect-browser';
```
jossmac, line 13: We include `@react-aria/utils`, you should be able to get this behaviour from the [platform](https://github.com/adobe/react-spectrum/blob/main/packages/%40react-aria/utils/src/platform.ts) utils.

packages/keystatic/src/component-blocks/cloud-image-preview.tsx
```diff
@@ -10,65 +10,189 @@ import { Icon } from '@keystar/ui/icon';
...
 
-  const dimensions = useImageDimensions(src);
+const browser = detect();
+const metaSymbol = browser?.os === 'Mac OS' ? '⌘' : 'Ctrl';
+const isMobile = browser?.os === 'iOS' || browser?.os === 'Android OS';
```
jossmac, line 17: Since this is used for language around pointer events, a more accurate condition might be:

```js
import { useMediaQuery } from '@keystar/ui/style';

let isTouchDevice = useMediaQuery('(pointer: coarse)');
```

packages/keystatic/src/component-blocks/cloud-image-preview.tsx
```diff
@@ -10,65 +10,189 @@ import { Icon } from '@keystar/ui/icon';
...
+              {isMobile ? 'Tap' : 'Click'} here to paste an Image URL
+            </Flex>
+          )}
+        </div>
+        {config.storage.kind === 'cloud' && (
+          <div>
+            <ActionButton
+              href={getProjectImagesURL(config.storage.project)}
+              target="_blank"
+              rel="noreferrer noopener"
```
jossmac, line 165: Nit: `noopener` is [implied](https://html.spec.whatwg.org/multipage/links.html#link-type-noreferrer) by `noreferrer`.

packages/keystatic/src/component-blocks/cloud-image-preview.tsx
```diff
@@ -10,65 +10,189 @@ import { Icon } from '@keystar/ui/icon';
...
+              borderRadius: '6px',
+              width: '98%',
+              height: '100%',
+            }}
+          />
+          {isFocused ? (
+            <Flex
+              backgroundColor="positive"
+              border="positive"
+              color="positive"
```
jossmac, line 140: The `color` prop is only valid on typographic primitives. If you're not seeing a type error, there's a bug.

packages/keystatic/src/component-blocks/cloud-image-preview.tsx
```diff
@@ -10,65 +10,189 @@ import { Icon } from '@keystar/ui/icon';
...
+              borderRadius="regular"
+              padding="regular"
+            >
+              {isMobile ? 'Tap' : `Press ${metaSymbol} + V`} to paste an Image
+              URL
+            </Flex>
+          ) : (
+            <Flex
+              border="neutral"
+              color="neutral"
```
jossmac, line 151: [Same here](https://github.com/Thinkmill/keystatic/pull/554/files#r1299449313).

packages/keystatic/src/component-blocks/cloud-image-preview.tsx
```diff
@@ -190,3 +300,8 @@ function getDimension(value: unknown) {
   if (typeof value === 'number') return value.toString();
   return '';
 }
+
+function getProjectImagesURL(projectConfig: string) {
+  const [team, project] = projectConfig.split('/');
+  return `https://keystatic.cloud/teams/${team}/project/${project}/images`;
```
jossmac, line 306: Should `https://keystatic.cloud` come from some constants file?

packages/keystatic/src/component-blocks/cloud-image-preview.tsx
```diff
@@ -10,65 +10,189 @@ import { Icon } from '@keystar/ui/icon';
...
+    <NotEditable>
+      <VStack gap="regular">
+        <div style={{ position: 'relative' }}>
+          <input
+            type="text"
+            value=""
+            onFocus={() => setIsFocused(true)}
+            onBlur={() => setIsFocused(false)}
+            onClick={() => {
+              document.execCommand('paste');
```
emmatown, line 116: What is this doing? based on trying it on iOS, that's not how it works there? and [according to MDN, it's not at all supported](https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand#:~:text=or%20insertion%20point.-,paste,at%20the%20insertion%20point%20(replaces%20current%20selection).%20Disabled%20for%20web%20content.,-redo)  (except in browser extensions it seems)

packages/keystatic/src/component-blocks/cloud-image-preview.tsx
JedWatson: There's a bug, I found this in the TS autocomplete so didn't think to reach for a `Text` component (also wondered why it had no effect, but figured I'd leave that for the review)

packages/keystatic/src/component-blocks/cloud-image-preview.tsx
JedWatson: Weirdly, it seemed to work on my version of iOS, I may have been in a "trusted environment" the way I was previewing the build? although that doesn't make sense.

Alternatively (based on my on-device testing) the newer API (I think I was using `navigator.clipboard.readText()`) seemed to work on Safari Desktop but not iOS, whereas `execCommand` seemed to work on both.

Nevermind, I think we're going in a different direction anyway.

packages/keystatic/src/component-blocks/cloud-image-preview.tsx
jossmac: Ugh, layout components are polymorphic with the `elementType` prop—I guess it's including the [legacy attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes)…

jossmac: Resolved in #563

JedWatson:
**Update**

After a design session today, I've updated this with new and simpler/nicer UX.

Now when you insert a Cloud Image component into a document, you get this placeholder:

image

(@jossmac this isn't an ideal design, but I could get the [DialogTrigger](https://keystar-ui.vercel.app/package/dialog/dialog-trigger) to work with anything _other_ than a `Button` or `ActionButton` so I've made do, you may have some suggestions?)

(also @emmatown may come up with a way to open this dialog from the start, before the placeholder is inserted, but that's basically a progressive enhancement since we'll still need to handle empty Markdoc tags)

When you click it, you get this dialog:

image

Pasting a valid Image URL reveals the rest of the form and auto-fills the width and height:

image

And finally you get the image preview, you can use the edit button on the right to change the alt/width/height (or replace the image) and the delete button to remove it from the document:

image

This is probably pretty close to ready from a UX perspective, we can tweak the workflow a bit more when we figure out how to really dial it in, but it's working pretty well for now.

One thing it doesn't handle yet is if you have an invalid image URL in the document _already_ (e.g if an image is deleted, or if you commit an invalid src directly into the codebase) but I think it's OK to handle that later.

packages/keystatic/src/component-blocks/cloud-image-preview.tsx
```diff
@@ -2,159 +2,410 @@
...
+            }}
+            value={state.src}
+            description={
+              image
+                ? undefined
+                : 'Copy an Image URL from the Image Library and paste into this field to insert it.'
+            }
+            endElement={
+              status === 'loading' ? (
+                <div
```
jossmac, line 177: Any reason this doesn't use a layout primitive?

```js
<Flex height="regular" width="regular" alignItems="center" justifyContent="center" />
```
```

File: packages/keystatic/src/component-blocks/cloud-image-preview.tsx [hunk 9]
```diff
@@ -2,159 +2,410 @@
...
+  onChange,
+  onRemove,
+}: {
+  onChange: (data: ImageData) => void;
+  onRemove: () => void;
+}) {
+  return (
+    <NotEditable>
+      <Flex gap="regular">
+        <Box flex="1">
```
jossmac, line 274: I think this is unnecessary, you should be able to move the flex prop to the `Button` e.g.

```jsx
<Flex>
  <DialogTrigger>
    <Button flex />
  </DialogTrigger>
</Flex>
```

packages/keystatic/src/component-blocks/cloud-image-preview.tsx
```diff
@@ -2,159 +2,410 @@
...
-                {dimensions
-                  ? `${dimensions.width} x ${dimensions.height}`
-                  : null}
-              </ActionButton>
-            </div>
-          ) : null}
+                W {image.width} x H {image.height}
+              </Text>
+            </Box>
+            <Box>
```
jossmac, line 345: What do these do?

packages/keystatic/src/component-blocks/cloud-image-preview.tsx
```diff
@@ -2,159 +2,410 @@
...
+    img.src = src;
+    return () => {
+      img.onload = null;
+    };
+  }, [src]);
+
+  return (
+    <Dialog>
+      <Heading>
+        <Flex alignItems="center" gap="regular">
```
jossmac, line 131: Why have this inside the heading?

jossmac:
> (@jossmac this isn't an ideal design, but I could get the [DialogTrigger](https://keystar-ui.vercel.app/package/dialog/dialog-trigger) to work with anything _other_ than a `Button` or `ActionButton` so I've made do, you may have some suggestions?)

This is by design, to stop people doing dodgy stuff.

packages/keystatic/src/component-blocks/cloud-image-preview.tsx
JedWatson: Good catch, it's leftover code from the [example](https://keystar-ui.vercel.app/package/dialog#slots)

packages/keystatic/src/component-blocks/cloud-image-preview.tsx
JedWatson: I can't use the right tokens for `height` and `width` props that I need here, they're giving me TypeScript errors

JedWatson: Aah, `element.regular`, got it :+1:

packages/keystatic/src/component-blocks/cloud-image-preview.tsx
JedWatson: Boxes (block) around the Text (inline) components so they get spaced out correctly by the VStack

jossmac (approved):

PR #562, opened 2023-08-21
Split view animation

jossmac:
Prevent undesirable panel animation during mount


JedWatson (approved):
Oh that is _much nicer_

PR #563, opened 2023-08-21
Forward ref legacy attributes

jossmac:
Omit legacy attributes from `forwardRefWithAs` utility. Avoid issues like:

```jsx
<Flex color={string | undefined} />
```

### DIFF
--- a/design-system/pkg/src/utils/ts/forwardRefWithAs.ts
+++ b/design-system/pkg/src/utils/ts/forwardRefWithAs.ts
@@ -13,13 +13,7 @@ export type PropsWithElementType<P = unknown> = P & {
 };
 
 // https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes
-type LegacyAttributes =
-  | 'background'
-  | 'bgColor'
-  | 'color'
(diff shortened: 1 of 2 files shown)


design-system/pkg/src/utils/ts/forwardRefWithAs.ts
```diff
@@ -12,19 +12,32 @@ export type PropsWithElementType<P = unknown> = P & {
...
   DefaultElementType extends HTMLTag
 > = (
-  props: BaseProps & {
-    // these types are less specific than they could be but for now
-    // typescript not being ridiculously slow is more important
-    // they could possibly get more specific later
-    elementType?: HTMLTag;
-    ref?: React.Ref<HTMLElement>;
-  } & Omit<AllHTMLAttributes<HTMLElement>, keyof BaseProps>
+  props: Omit<
```
jossmac, line 30: I switched the order here so omission doesn't conflict with `BaseProps` where there's overlap.

JedWatson (approved):

emmatown (approved):

PR #568, opened 2023-08-22
Fix Firefox bug cleaning up cloud image components

JedWatson:
This fixes a bug where Firefox doesn't handle Slate focus as expected, and leaves the image placeholder.

It also adds some defensive text and a handler so you can clearly click on an image placeholder to bring up the dialog if you somehow end up with an invalid tag in your Markdoc document.

(and I removed the "no alt text" state, not a fan tbh)

@jossmac over to you for review, there may be better ways of handling this generally but this does make Firefox more robust for now.


JedWatson:
(we probably have enough change sets related to this field, it'll get released anyway so I don't see a lot of value in specifically calling this fix out in the changelog)

jossmac (approved):

PR #571, opened 2023-08-22
Remove status column in collection view in local mode

JedWatson:
I noticed that since the latest UI refresh, we're showing "unchanged" status for all the collection items in the list view when in local mode (i.e `{ storage: { mode: 'local' } }`)

That's really misleading, so this PR fixes it.

tbh not my finest work, the layout (esp. with row columns) is pretty finicky but I couldn't pull the cells out into a separate component without messing around with types (i.e the type of `item` is just inferred from the scope, so specifying it as a prop type in another component is beyond me but I'd love to see how it's done)

I also added a flex container to the label cells matching the height of the `StatusLight` component because otherwise the rows squish up and it's not my intention that the layout actually changes in local mode.

So, happy to get advice on how to structure this better, but it's definitely an improvement to not show invalid status now:
image

(on another note, I found [NodeGit](https://www.nodegit.org) and we ... could actually pull the changed status for local mode, but that feels like an adventure for another time)

### DIFF
--- a/packages/keystatic/src/app/CollectionPage.tsx
+++ b/packages/keystatic/src/app/CollectionPage.tsx
@@ -37,6 +37,7 @@ import { useTree, TreeData } from './shell/data';
 import { PageRoot, PageHeader } from './shell/page';
 import { getCollectionPath, getEntriesInCollectionWithTreeKey } from './utils';
 import { notFound } from './not-found';
+import { Flex } from '@keystar/ui/layout';
 
 type CollectionPageProps = {
   collection: string;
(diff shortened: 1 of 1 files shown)


JedWatson:
p.s not sure this needs a changeset, it's fixing a bug that's been introduced since the last release. the package is being released anyway and I'm not sure it would make sense in the changelog

jossmac:
> (on another note, I found [NodeGit](https://www.nodegit.org) and we ... could actually pull the changed status for local mode, but that feels like an adventure for another time)

Oh, that could be cool

jossmac (approved):
Added a little fix so this works for all storage modes.

emmatown (approved):

PR #586, opened 2023-08-24
Export CloudImageProps

JedWatson:
I was watching @lukebennett88 implement a `cloudImage` component block renderer in his site, and he had to guess/recreate this type in his front-end.

Seems better for us to export it and allow people to import the same definition we're using internally.

(I also consolidated the duplicate types between the component block and the field, since other code is also being shared)

### DIFF
--- a/packages/keystatic/src/component-blocks/cloud-image.tsx
+++ b/packages/keystatic/src/component-blocks/cloud-image.tsx
@@ -2,6 +2,8 @@ import { component } from '..';
 import { CloudImagePreview } from './cloud-image-preview';
 import { cloudImageSchema } from './cloud-image-schema';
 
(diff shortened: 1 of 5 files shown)


emmatown (approved):

packages/keystatic/src/component-blocks/cloud-image.tsx
```diff
@@ -2,6 +2,8 @@ import { component } from '..';
 import { CloudImagePreview } from './cloud-image-preview';
 import { cloudImageSchema } from './cloud-image-schema';
 
+export type { CloudImageProps } from './cloud-image-preview';
```
emmatown, line 5: Needs to be re-exported from `packages/keystatic/src/component-blocks/index.tsx`

PR #590, opened 2023-08-24
Website component blocks cleanup

JedWatson:
This cleans up our docs website config considerably.

* Component blocks are defined in their own files now, which makes the config easier to understand
* The old cloud images field has been removed the new one is compatible with the existing images (and I've updated the tags to use the new `src` prop, and integers for `width` and `height`)
* The `field-demo` tag casing is consistent with the other tags now
* The previews correctly use `NotEditable`

### DIFF
--- a/docs/src/components/document-renderer.tsx
+++ b/docs/src/components/document-renderer.tsx
@@ -162,7 +162,6 @@ const componentBlockRenderers: InferRenderersForComponentBlocks<
         src={src}
         height={height ?? undefined}
         width={width ?? imgMaxWidthPx}
-        sizes={`(max-width: 375px) 375px, ${imgMaxWidthPx}px`}
         className="rounded-lg my-2"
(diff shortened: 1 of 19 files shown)


docs/src/components/document-renderer.tsx
```diff
@@ -186,6 +162,7 @@ const componentBlockRenderers: InferRenderersForComponentBlocks<
         src={src}
         height={height ?? undefined}
         width={width ?? imgMaxWidthPx}
+        sizes={`(max-width: 375px) 375px, ${imgMaxWidthPx}px`}
         className="rounded-lg my-2"
```
emmatown, line 166: There's no `srcset` here so replacing the components feels premature?

emmatown (changes requested):

emmatown (approved):

PR #591, opened 2023-08-24
Cloud Image Component Fixes

JedWatson:
* Fixes a bug where opening the edit modal for an image that's already in the document would not actually populate the form
* Don't append the transform strings unless the image is being hosted on `keystatic.net` (this was causing problems in the docs site with images coming from a different service)
* Minor UI tweaks

### DIFF
--- a/packages/keystatic/src/component-blocks/cloud-image-preview.tsx
+++ b/packages/keystatic/src/component-blocks/cloud-image-preview.tsx
@@ -279,9 +279,10 @@ function Placeholder(props: {
         gap="regular"
         height="element.large"
         paddingX="large"
+        onClick={() => state.open()}
       >
         <Icon src={imageIcon} />
-        <Text>Cloud image (click to configure)</Text>
+        <Text>Cloud image{state.isOpen ? '' : '(click to configure)'}</Text>
       </Flex>
       <DialogContainer onDismiss={closeAndCleanup}>
         {state.isOpen && (
--- a/packages/keystatic/src/component-blocks/cloud-image-preview.tsx
+++ b/packages/keystatic/src/component-blocks/cloud-image-preview.tsx
@@ -415,7 +415,7 @@ function imageWithTransforms(options: {
 }) {
   let { fit = 'scale-down', source, height, width } = options;
 
-  if (!/https?:\/\/[^\.]+.keystatic.net/.test(source)) {
+  if (!/^https?:\/\/[^\.]+\.keystatic\.net/.test(source)) {
     return source;
   }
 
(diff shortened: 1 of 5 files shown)


packages/keystatic/src/component-blocks/cloud-image-preview.tsx
```diff
@@ -281,7 +281,7 @@ function Placeholder(props: {
         paddingX="large"
       >
         <Icon src={imageIcon} />
-        <Text>Cloud image, awaiting configuration…</Text>
+        <Text>Cloud image (click to configure)</Text>
```
jossmac, line 284: The `Placeholder` won't render if there's an image source. How could a user ever click this non-interactive element?

packages/keystatic/src/component-blocks/cloud-image-preview.tsx
```diff
@@ -418,6 +414,10 @@ function imageWithTransforms(options: {
 }) {
   let { fit = 'scale-down', source, height, width } = options;
 
+  if (!/https?:\/\/[^\.].keystatic.net/.test(source)) {
```
emmatown, line 417: This would make testing and custom domains quite impractical so i'm not very keen on it? separately, it wouldn't ever work right:
```suggestion
  if (!/https?:\/\/[^\.]+.keystatic.net/.test(source)) {
```

emmatown (changes requested):

packages/keystatic/src/component-blocks/cloud-image-preview.tsx
JedWatson: Yeah, good catch on the regex, thanks

Agreed, was thinking about custom domains when I write it and I've got ideas on how we can make this better later (before we get to custom domains even) but for now it was causing a 404 on images from a different service so I think it's worth putting in for now.

packages/keystatic/src/component-blocks/cloud-image-preview.tsx
jossmac, line 284: The `Placeholder` won't render if there's an image source. How could a user ever click this non-interactive element?

jossmac (changes requested):

packages/keystatic/src/component-blocks/cloud-image-preview.tsx
```diff
@@ -418,6 +415,10 @@ function imageWithTransforms(options: {
 }) {
   let { fit = 'scale-down', source, height, width } = options;
 
+  if (!/https?:\/\/[^\.]+.keystatic.net/.test(source)) {
```
emmatown, line 418:
```suggestion
  if (!/^https?:\/\/[^\.]+\.keystatic\.net/.test(source)) {
```

emmatown (approved):

PR #596, opened 2023-08-28
KEY-234: Port showcase to Docs site

simonswiss:
BIG SCARY PR!

There are many files changed. Here's the gist of it:

- Ports the `showcase` website (currently a standalone Astro site) into the `docs` Next.js site
- Implements the new Figma design
- Reworks the colors throughout the Keystatic docs with the Keystart UI Tokens (overwrite default colors and make sure the UI doesn't use any default color)
- Flows element of the brand refresh (colors, font weight, badges etc) across the docs pages beyond just the showcase.
- Change the Tailwind config to a TS file
- Use Keystatic Cloud for the showcase images

Not ideal to review the PR, but a bunch of files only changed because I setup the `prettier-plugin-tailwindcss` properly on the docs project — which sorts Tailwind classes consistently.


docs/src/content/projects/better-dev-screencasts.mdoc
```diff
@@ -0,0 +1,64 @@
...
+title: Better Dev Screencasts
+type: production
+url: https://betterdevscreencasts.com
+repoUrl: https://github.com/simonswiss/better-dev
+summary: >-
+  A collection of resources to help developers create better screencasts, and
+  level-up their overall video content creation game.
+coverImage:
+  src: >-
+    https://thinkmill-labs.keystatic.net/keystatic-site/images/t1i890vdj0cf/keystatic-site
```
BarnabyBishop, line 11: This seems to be a keystatic docs screenshot instead of a Better Dev Screencasts 
(I was snooping on your docs site updates and notice this, looking good!)

simonswiss: Ooh — good catch! And thank you 🙏🏼 

Updated 👍

docs/src/app/(public)/layout.tsx
```diff
@@ -10,7 +10,7 @@ export default async function RootLayout({
   const navigationMap = await getNavigationMap();
 
   return (
-    <div className="min-h-screen flex flex-col">
+    <div className="flex min-h-screen flex-col text-slate-12 antialiased">
```
simonswiss, line 13: The `antialiased` class here is responsible for the whole site feeling a bit "thinner".

Uploading antialiased.mov…

simonswiss:
Just increased the surface area of the PR 😅 by phasing out all the images on the `keystatic.io` domain — now all images in `docs` are hosted on the Thinkmill Labs Keystatic Cloud account 👍

JedWatson (approved):

PR #614, opened 2023-08-31
Cloud image constrain proportions

jossmac:
Support constrained proportions on cloud image block and field. Refactor `Tooltip` styles to allow consumer overrides via style props.


emmatown (approved):
We'll want to get this into the cloud image field as well but this looks good

PR #618, opened 2023-09-05
Improve bundle size with reader API in React server component environments

### LINKED ISSUE #453: createReader adds ~500kB to bundle size with /app directory
Steps to reproduce:
1. Create a new project with `npm create @keystatic@latest`
2. Build it (using yarn and nextjs)

Now `/` and `/[slug]` have a bundle size of over ~580kB.

It seems like `createReader` gets bundled into the client bundle, because when removing it it has the expected size of about ~70kB

Versions:
"@keystatic/core": "^0.0.109",
"@keystatic/next": "^0.0.11",
"next": "^13.4.4",

macOS 13.4 (22F66)
node v18.15.0

### LINKED ISSUE #601: Importing @keystatic/core will increase the bundle size by 500kB+
## Description

After importing @keystatic/core into the Next.js application directory, any page that references keystatic will load the Keystatic UI bundle js and dependencies.

bundle size image

After a brief investigation, I found that there are a lot of `import 'module'` statements in the `keystatic-core.{esm/cjs}.js` file, and Next.js is not able to do tree shaking on them, which results in a larger bundle output by Next.js.

keystatic-core js file image

## Reproduction steps

1. Clone this repository
2. Go to templates/nextjs folder
3. Run npm install
4. Run npm run build
5. View console output

## My Project

When I remove the Keystatic UI page from the app directory, the First Load JS of the page is still 580kB+, and if I remove `import { something } from "@keystatic/core"` from the page, the First Load JS of the page will be the normal size (142kB+).

Abnormal size:

image

image

Normal size:

image

image

## Versions

Template project:

@keystatic/core@0.1.1
@keystatic/next@1.0.0
@types/react@18.2.21
next@13.4.19
react-dom@18.2.0
react@18.2.0
typescript@5.2.2

My project:

@keystatic/core 0.1.1
@keystatic/next 1.0.0
@types/react 18.2.21
next 13.4.19
react-dom 18.2.0
react 18.2.0
typescript 5.2.2

emmatown:
Fixes #453 
Fixes #601

This replaces `'use client'`s in UI modules imported by code that's needed for the reader API with using the `package.json#imports` field to map those UI modules to empty implementations in `react-server` environments so that Next will never bundle those UI modules into pages using the reader API

As an example of the improvement, in the `dev-projects/next-app`, using the reader API made the first load JS `600 kB`, after these changes, it's `77.7 kB` which seems to be the baseline with Next. (testing with the custom `Note` component block that's there disabled in both cases)


packages/keystatic/package.json
```diff
@@ -284,6 +284,18 @@
...
+    },
+    "#field-ui/*": {
+      "react-server": "./src/form/fields/empty-field-ui.tsx",
+      "default": "./src/form/fields/*/ui.tsx"
+    },
+    "#component-block-primitives": {
+      "react-server": "./src/form/fields/document/DocumentEditor/primitives/blank-for-react-server.tsx",
+      "default": "./src/form/fields/document/DocumentEditor/primitives/index.tsx"
+    },
+    "#cloud-image-preview": {
```
JedWatson, line 296: this feels ... moderately unsustainable, if we end up with many more component blocks? could/should we organise this differently, like have an `component-blocks` import? or is this just private API and it doesn't matter, we can refactor it later if that's an issue?

emmatown: > is this just private API and it doesn't matter, we can refactor it later if that's an issue?

This, all of the `imports` field is entirely private

JedWatson (approved):

PR #620, opened 2023-09-05
Fix astro integration

### LINKED ISSUE #580: Error: "lodash" does not provide an export member "get"
I was interest in learning more about Keystatic, so I gave it a go with the "astro-content" under dev-projects (mainly since it differed from the docs). I apologize if I shouldn't have been looking there, but I just want to point out the following error if you tried to use it:

Uncaught (in promise) SyntaxError: The requested module '/node_modules/.pnpm/lodash@4.17.21/node_modules/lodash/lodash.js?v=aa3ff6fd' does not provide an export named 'get' (at keystar-ui-style.esm.js?v=aa3ff6fd:7:10)

emmatown:
This generates a file into `.astro` (since that will already be gitignored) with imports to Keystatic packages and tells Vite to look at it for discovering dependencies to pre-bundle which fixes the problem of getting errors from Vite trying to load CommonJS modules as ESM

I'm not going to update the template to use this just yet, I'd like to get this released first to make sure that it definitely works first (though from the snapshot release, it seems to)

Snapshot release for testing: `@keystatic/astro@0.0.0-test-20230907032149`, https://github.com/Thinkmill/keystatic/actions/runs/6105022202/job/16567883939#step:7:33

Fixes  #580

### DIFF
--- a/.changeset/odd-paws-exist.md
+++ b/.changeset/odd-paws-exist.md
@@ -1,0 +1,5 @@
+---
+'@keystatic/astro': patch
+---
+
+Fix `@keystatic/astro` integration
--- a/packages/astro/src/index.ts
+++ b/packages/astro/src/index.ts
@@ -29,10 +29,14 @@ export default function keystatic(): AstroIntegration {
             entries: ['keystatic.config.*', '.astro/keystatic-imports.js'],
           },
         };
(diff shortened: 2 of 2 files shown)


JedWatson (approved):
Wow. If this works, I love it*

PR #625, opened 2023-09-06
Astro docs

valery-tm:
Updated Astro related docs with some extra steps and fixed the port number.


JedWatson (changes requested):
Looks good!

Please add a changeset for the update to `@keystatic/create` though

JedWatson (approved):

PR #628, opened 2023-09-07
Replace manually created pages for Keystatic with Astro integration that injects routes

emmatown:
#620 seems to work okay, so looks like we can use it here


JedWatson (approved):
🥳

PR #635, opened 2023-09-09
Update astro-disable-admin-ui-in-production.mdoc

barelyhuman:
The API route disable in production code seems to be invalid. 

Correcting it by adding a wrapper around the `makeHandler` in the call. If there's a better API that does this then please let me know and I'll make the needed corrections


florian-lefebvre (approved):
LGTM, I actually didn't tested with api routes so I'm not surprised it does not work

emmatown (approved):

PR #644, opened 2023-09-18
Fields editing: modal

valery-tm:
Editing of the fields should happen in a modal instead of in-document.

### DIFF
--- a/packages/keystatic/src/form/api.tsx
+++ b/packages/keystatic/src/form/api.tsx
@@ -292,7 +292,6 @@ export type ComponentBlock<
       chromeless?: false;
       toolbar?: (props: {
         props: Record<string, any>;
-        onShowEditMode(): void;
         onRemove(): void;
         isValid: boolean;
       }) => ReactElement;
--- a/packages/keystatic/src/form/api.tsx
+++ b/packages/keystatic/src/form/api.tsx
@@ -292,6 +292,7 @@ export type ComponentBlock<
       chromeless?: false;
       toolbar?: (props: {
         props: Record<string, any>;
+        onShowEditMode(): void;
         onRemove(): void;
         isValid: boolean;
       }) => ReactElement;
--- a/packages/keystatic/src/form/fields/document/DocumentEditor/component-blocks/chromeful-element.tsx
+++ b/packages/keystatic/src/form/fields/document/DocumentEditor/component-blocks/chromeful-element.tsx
@@ -8,11 +8,12 @@ import {
   forwardRef,
   Ref,
   useId,
+  useCallback,
 } from 'react';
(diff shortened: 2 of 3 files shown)


packages/keystatic/src/form/api.tsx
```diff
@@ -292,7 +292,6 @@ export type ComponentBlock<
       chromeless?: false;
       toolbar?: (props: {
         props: Record<string, any>;
-        onShowEditMode(): void;
```
emmatown, line 295: Could you add this prop back since we probably don't want to do a breaking change for this? You'll probably want to use the [`DialogContainer`](https://keystar-ui.vercel.app/package/dialog/dialog-container) rather than `DialogTrigger` component so that you can manage the state yourself

packages/keystatic/src/form/fields/document/DocumentEditor/component-blocks/chromeful-element.tsx
```diff
@@ -190,38 +181,51 @@ function DefaultToolbarWithChrome({
...
 }: {
   props: GenericPreviewProps<NonChildFieldComponentSchema, unknown>;
   onClose(): void;
-  isValid: boolean;
 }) {
+  const stringFormatter = useLocalizedStringFormatter(l10nMessages);
+  const formId = useId();
   const [forceValidation, setForceValidation] = useState(false);
+  const val: unknown[] = previewPropsToValue(props as any);
+  const [state, setState] = useState(val);
```
emmatown, line 192:
```suggestion
  const [state, setState] = useState(() => previewPropsToValue(props as any) as unknown);
```

Just since it only needs to happen on mount

packages/keystatic/src/form/fields/document/DocumentEditor/component-blocks/chromeful-element.tsx
```diff
@@ -167,9 +153,14 @@ function DefaultToolbarWithChrome({
     <NotEditable>
       <Flex direction="column" gap="medium">
         <Flex alignItems="center" gap="regular" style={{ userSelect: 'none' }}>
-          <ActionButton onPress={() => onShowEditMode()}>
-            {stringFormatter.format('edit')}
-          </ActionButton>
+          <DialogTrigger>
+            <ActionButton>{stringFormatter.format('edit')}</ActionButton>
+            {close => (
+              <Dialog onDismiss={close}>
```
emmatown, line 159: Assuming the dialog rendering moves into `ChromefulComponentBlockElement`, could you also add a heading to the modal, probably like `Edit {componentBlock.label}`?

emmatown (commented):
Looks great! 🎉

emmatown (approved):
🚢

PR #648, opened 2023-09-20
Allow cloud auth in local mode when configured

emmatown:


packages/keystatic/src/app/auth.ts
```diff
@@ -20,23 +21,32 @@ export function getSyncAuth(config: Config) {
...
+  let tokenData;
+  try {
+    tokenData = storedTokenSchema.parse(JSON.parse(unparsedTokenData!));
+  } catch (err) {
+    return null;
+  }
+  if (
+    !tokenData ||
+    tokenData.validUntil < new Date() ||
+    tokenData.project !== config.cloud.project
```
emmatown, line 43: This is so that we don't try using a token that is for a different project than it was created for which is pretty likely to happen running different projects locally over time on the same port

PR #656, opened 2023-09-28
og image update

chrisdale-io:
1. Updating default og image (change of image path)
2. Simplifying the previous code implementation in the route file. Replacing complex layering of tiled background texture image over background svg, to simply a background png.

### DIFF
--- a/docs/src/app/og/route.tsx
+++ b/docs/src/app/og/route.tsx
@@ -30,11 +30,7 @@ export async function GET(req: NextRequest) {
           height: '100%',
           width: '100%',
           backgroundImage:
-            'url(https://thinkmill-labs.keystatic.net/keystatic-site/images/km9uvzfb8fqo/og-image-bg)',
-          backgroundRepeat: 'no-repeat',
-          backgroundAttachment: 'fixed',
-          backgroundSize: 'cover',
-          backgroundColor: '#F9F9F8',
+            'url(https://thinkmill-labs.keystatic.net/keystatic-site/images/km9uvzfb8fqo/og-image-bg?format=png&width=1200)',
         }}
       >
         <div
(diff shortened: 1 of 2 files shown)


docs/src/app/og/route.tsx
```diff
@@ -29,104 +29,14 @@ export async function GET(req: NextRequest) {
           display: 'flex',
           height: '100%',
           width: '100%',
+          backgroundImage:
+            'url(https://thinkmill-labs.keystatic.net/keystatic-site/images/km9uvzfb8fqo/og-image-bg)',
```
emmatown, line 33: This background image doesn't seem to be used? so the background is just the colour? I did some investigation and it seems like it's because of https://github.com/vercel/satori/issues/273 and Keystatic Cloud serving webp by default, you should be able to fix that by using `https://thinkmill-labs.keystatic.net/keystatic-site/images/km9uvzfb8fqo/og-image-bg?format=png` (from testing that seemed to solve the error from webp but the image still wasn't actually being displayed)

chrisdale-io: Hey Emma, thanks for looking into this. I had a bit of a look and test a few ideas from what you've come back with, with no success. I'm not sure why it's not loading up the background image, as it seemed to work fine from the implementation before the brand update work!?

The previous code for the background image in the original version was this:
`backgroundImage: 'url(https://thinkmill-labs.keystatic.net/keystatic-site/images/nt24l5xes6jj/open-graph-bg)',`

emmatown: From checking out the commit before the brand updates (183e1293df2871f19e1239965a19f29e2b700846) and trying it now, it seems as though it broke. I'm guessing that would have broke when we added image transforms to `*.keystatic.net` 😬  (though adding `?format=png` to the url in 183e1293df2871f19e1239965a19f29e2b700846 does seem to work)

emmatown: Just pushed a commit that seems to fix it, it seems like having an image bigger than the image that's being created made something decide to not render it so resizing the image to be the same size as the output worked

emmatown (approved):

PR #665, opened 2023-10-02
Fix(url): field validation

### LINKED ISSUE #661: Field URL Validation not working
# Config:
image

# Not Working:
image

ArnavK-09:
Resolves #661 
Fixes #661


ArnavK-09:
@emmatown ? It's ruining my ux in production site

JedWatson:
Hey @ArnavK-09 

@emmatown is on holiday this week, but I've tried to reproduce the problem you're reporting and can't quite understand how it's not working for you (and also, how this fixes it)

I'm happy to push through a fix for you while she's away, but will need some more information to ensure this is the correct change.

The current validation logic is intended to:

* Ensure that if there **is** a value provided, it is actually a valid URL
* Only if `isRequired === true`, also ensure that a value (not an empty string) has been provided

When I add a field with this config to my test project:

```
url: fields.url({ label: 'URL' }),
```

... it doesn't seem to cause a problem saving an entry with no value in the field (which is how I interpret your fix to work)

Setting the option to false, like this, also works:

```
url: fields.url({ label: 'URL', validation: { isRequired: false } }),
```

The way I read the change you're proposing here, format validation would only happen if the field is required, which is not the intended behaviour. If you want to allow any format of value in the field, you could just use `fields.text`

Can you help clarify what behaviour you're seeing that's different, and how this fix resolves the issue?

Thanks

ArnavK-09:
Hey @JedWatson 
> **Thanks for looking into the matter, actually the problem triggers on creating new post.**, Kindly checkout  Some screenshots below

---

# The problem 

https://github.com/Thinkmill/keystatic/assets/69188140/a186988a-8c01-47f0-b114-414d86a1505b

# My Default Config 

boot 10_5_2023 9_48_17 PM

# Even my I don't pass any validation prop

https://github.com/Thinkmill/keystatic/assets/69188140/4cff9c1c-283e-4d49-a44b-b4b6faa07f26

---

**I request you to kindly fix #661, and it would be alright if this pull req doesn't fix the issue actually I haven't tested the changes, but please positively fix with any other change**

> ✨ **Regards**

ArnavK-09:
Maybe !isRequired validation is extra in my pull req I guess so, since on init the value passed is '' not ` null `, ig

emmatown (approved):

PR #679, opened 2023-10-18
Update urql

emmatown:
I noticed that we were doing the same graphql query twice in cloud mode for some reason, upgrading urql seems to fix it. (I think it's probably to do with `@urql/exchange-persisted`)


packages/keystatic/src/app/provider.tsx
```diff
@@ -48,7 +47,6 @@ export function createUrqlClient(config: Config): Client {
         : `${KEYSTATIC_CLOUD_API_URL}/v1/github/graphql`,
     requestPolicy: 'cache-and-network',
     exchanges: [
-      dedupExchange,
```
emmatown, line 51: Removed this because it's deprecated and the same functionality is built-in now

jossmac (approved):

PR #687, opened 2023-10-23
Add `createGitHubReader`

emmatown:


packages/keystatic/src/reader/generic.ts
```diff
@@ -0,0 +1,445 @@
...
+    readOrThrow: async (...args) => {
+      const entry = await (read as any)(...args);
+      if (entry === null) {
+        throw new Error(
+          `Entry "${args[0]}" not found in collection "${collection}"`
+        );
+      }
+      return entry;
+    },
+    // TODO: this could drop the fs.stat call that list does for each item
```
JedWatson, line 298: This doesn't seem relevant to the code below?

JedWatson (approved):

packages/keystatic/src/reader/generic.ts
emmatown: It is, `list` does checks that aren't really necessary since we immediately read the item in `all`

PR #690, opened 2023-10-24
`toolbarIcon` option in component blocks

emmatown:
keystatic with the cursor over a button in the editor toolbar with an image icon and a tooltip saying cloud image

### DIFF
--- a/packages/keystatic/src/form/api.tsx
+++ b/packages/keystatic/src/form/api.tsx
@@ -280,6 +280,7 @@ export type ComponentBlock<
   preview: (props: any) => ReactElement | null;
   schema: Fields;
   label: string;
+  toolbarIcon?: ReactElement;
 } & (
   | {
       chromeless: true;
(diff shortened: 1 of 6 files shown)


packages/keystatic/src/form/api.tsx
```diff
@@ -451,6 +452,8 @@ export function component<
     schema: Schema;
     /** The label to show in the insert menu and chrome around the block if chromeless is false */
     label: string;
+    /** An icon to show in the toolbar for this component block. Component blocks with `toolbarIcon` are shown in the toolbar directly instead of the insert menu */
```
jossmac, line 455: > instead of the insert menu

Are component blocks with `toolbarIcon` omitted from the insert menu?

jossmac (approved):

packages/keystatic/src/form/api.tsx
emmatown: Yes, it's slightly confusing since by "insert menu" here, I mean menu on the plus button in the toolbar, not the `/` menu, they still appear in the `/` menu (as do many other things that aren't in the other menu in the toolbar)

PR #691, opened 2023-10-24
Table view cleanup

jossmac:
See commits. Added lines are mostly stories.

App change: support `mod+f` on collection page.

### DIFF
--- a/packages/keystatic/src/app/CollectionPage.tsx
+++ b/packages/keystatic/src/app/CollectionPage.tsx
@@ -89,10 +89,16 @@ function CollectionPageHeader(props: {
     setSearchVisible(isAboveMobile);
   }, [isAboveMobile]);
 
-  // the entries are presented in a virtualized table view, so we have to
-  // replace the default (e.g. ctrl+f) browser search behaviour
+  // entries are presented in a virtualized table view, so we replace the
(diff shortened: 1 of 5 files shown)


packages/keystatic/src/app/CollectionPage.tsx
```diff
@@ -88,6 +89,19 @@ function CollectionPageHeader(props: {
     setSearchVisible(isAboveMobile);
   }, [isAboveMobile]);
 
+  // the entries are presented in a virtualized table view, so we have to
+  // replace the default (e.g. ctrl+f) browser search behaviour
+  useEffect(() => {
+    const listener = (event: KeyboardEvent) => {
+      if (isHotkey('mod+f', event)) {
+        event.preventDefault();
+        searchRef.current?.select();
```
emmatown, line 98: Do we want to avoid prevent defaulting if the search input is already focused so people can mod+f twice if they want to open the browser search for whatever reason?

jossmac: Yeah, that sounds reasonable.

- [ ] allow browser-find when hotkey repeated whilst search focused

emmatown (approved):

PR #694, opened 2023-10-25
Image library button

jossmac:
Move image library button from dashboard to app topbar.


packages/keystatic/src/app/shell/topbar.tsx
```diff
@@ -109,6 +111,7 @@ function CloudHeader({ config }: { config: CloudConfig }) {
       <BranchPicker />
       <GitMenu />
       <Box flex="1" />
+      <ImageLibraryButton />
```
emmatown, line 114: We should also have this in the `LocalHeader`

jossmac: Good catch, thanks!

emmatown (approved):

PR #698, opened 2023-10-25
Docs: installation-astro.mdoc: Add file names to code blocks

tordans:
All other code blocks have a file name, which makes it a lot easier to follow along. Those two did not. This PR adds the missing file name suggestions.


simonswiss (approved):
Nice one, thanks for that!

PR #699, opened 2023-10-25
Docs: installation-astro.mdoc: Add hint at rendering collections

tordans:
I paused when I saw that the guide only renders one file. This PR links over to the astro docs which will lead one to create the rendering vor all posts.

I thought about adding a section with code right here, but the having this collection rendering makes most sense when one also has a layout file and that is too much to introduce right here, IMO. But users can figure this out in the great Astro guides easily.


simonswiss:
Yep, agree this is useful but adding a code example would be too much, and deriving from the simplicity of this guide.

Thanks for this addition 👍

docs/src/content/pages/installation-astro.mdoc
```diff
@@ -176,6 +176,8 @@ const { Content } = await post.render()
 </main>
 ```
 
+Now you can follow the Astro Guide to [generate pages for all entries of a collection](https://docs.astro.build/de/tutorials/add-content-collections/#generate-pages-from-a-collection).
```
simonswiss, line 179:
```suggestion
If you want to generate pages for all entries of a collection, [this Astro Guide](https://docs.astro.build/de/tutorials/add-content-collections/#generate-pages-from-a-collection) will help.
```

simonswiss (approved):

PR #700, opened 2023-10-25
Docs: Add hint to changelog to readme

tordans:
I was looking for the changelog on the Repo page in Github and could not find it easily. ~~It is "hidden" in the docs in this Repo, which is a bit unusual IMO. This add the search term to the readme so people who cmd+f for "changelog" will find it faster. (I did find it using the file search…).~~

Update: Sorry, this is wrong. There are multiple changelogs.


tordans:
Sorry, I see now that in this multi repo there are all kind of changelogs. Will close this…

FYI, I was trying to figure out why https://github.com/Thinkmill/keystatic/pull/621 is not how my newly generated side was set up.

PR #726, opened 2023-11-01
Action group links

jossmac:
Support link as `ActionGroup` items.

### DIFF
--- a/design-system/pkg/src/style/classNames.ts
+++ b/design-system/pkg/src/style/classNames.ts
@@ -45,23 +45,23 @@ export class ClassList<ElementName extends string> {
     );
   }
 
-  get(element: ElementName | 'root') {
-    if (element === 'root') {
+  element(name: ElementName | 'root') {
+    if (name === 'root') {
       return this.root;
     }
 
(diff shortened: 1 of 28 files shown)


design-system/pkg/src/style/classNames.ts
```diff
@@ -23,36 +23,54 @@ export function classNames(...inputs: ClassNamesArg[]) {
...
 
 /**
- * A thin wrapper around Map to manage component class names. Keys are prefixed
- * with the component name and Voussoir's prefix.
+ * A ClassList organises component class names with appropriate prefixes,
+ * offering strongly-typed methods for declaration in JSX and styles.
  */
-export class ClassList extends Map {
-  prefix: string;
+export class ClassList<T extends readonly string[], K extends T[number]> {
```
jossmac, line 29: Might be worth a chat about the intent/use of this.

emmatown (approved):

jossmac:
The `ClassList` fixes got out of hand. I'll leave it where it is now and follow up in a dedicated PR.

PR #728, opened 2023-11-02
Use minimal router in `@keystatic/core` over different ones for each framework

emmatown:
This improves performance of navigating around Keystatic by avoiding using different routers for each framework and instead has some minimal pushState & popstate handling in Keystatic itself. It's particularly good in the Next app directory because it avoids fetching the server component payload on each payload (that takes e.g. 300ms for me on https://keystatic.com/keystatic). It's a breaking change for the different framework packages since they need the newer version of `@keystatic/core`.


jossmac (approved):
Much simpler!

PR #732, opened 2023-11-03
Don't pin peer dependency to exact next.js version

stefanprobst:
don't pin next.js peer dependency to exact version to avoid errors like:

```
 WARN  Issues with peer dependencies found
.
└─┬ @keystatic/core 0.1.9
  └─┬ @keystar/ui 0.4.0
    └── ✕ unmet peer next@14.0.0: found 14.0.1

```

### DIFF
--- a/design-system/pkg/package.json
+++ b/design-system/pkg/package.json
@@ -6241,7 +6241,7 @@
     "tsx": "^3.8.0"
   },
   "peerDependencies": {
-    "next": "^14.0.0",
+    "next": "13 || 14",
     "react": "^18.2.0",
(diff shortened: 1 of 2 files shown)


design-system/pkg/package.json
```diff
@@ -6241,7 +6241,7 @@
     "tsx": "^3.8.0"
   },
   "peerDependencies": {
-    "next": "14.0.0",
+    "next": "^14.0.0",
```
emmatown, line 6244:
```suggestion
    "next": "13 || 14",
```

emmatown (approved):
Thanks!

PR #739, opened 2023-11-08
Interface branding

jossmac:
New config `ui` supports branding options.

### DIFF
--- a/packages/keystatic/src/app/shell/common.tsx
+++ b/packages/keystatic/src/app/shell/common.tsx
@@ -6,6 +6,7 @@ import {
   useMediaQuery,
 } from '@keystar/ui/style';
 
+import { serializeRepoConfig } from '../repo-config';
 import { useConfig } from './context';
 
 export function useBrand() {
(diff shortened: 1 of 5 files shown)


packages/keystatic/src/app/shell/common.tsx
```diff
@@ -1,6 +1,36 @@
...
-export function ZapLogo() {
+import { useConfig } from './context';
+
+export function useBrand() {
+  let { colorScheme } = useProvider();
+  let config = useConfig();
+  let prefersDark = useMediaQuery('(prefers-color-scheme: dark)');
+
+  let brandMark = <ZapLogo />;
+  let brandName = config.cloud ? config.cloud.project : 'Keystatic';
```
emmatown, line 17: This means we're not showing the GitHub repo when in GitHub mode?

jossmac: TBH I've always found it strange that was so prominent; very easy to misunderstand and click assuming it will take the user "home". The repo is still available from the git menu:

image

emmatown: Sure, I don't mind about it not being a link but it seems a bit weird to show the cloud project key when in cloud mode but not the repo in github mode?

jossmac: Oh right, you mean default it? Yeah, for sure

emmatown (approved):

PR #751, opened 2023-11-13
[Feat] - Hide branches management UI

Barruchou:
Following this discussion on Discord:

https://discord.com/channels/1139017922073866332/1139017922895937598/1170003038942003210

Here is a PR to allow you to hide branch management elements from the interface.

### DIFF
--- a/packages/keystatic/src/app/dashboard/DashboardPage.tsx
+++ b/packages/keystatic/src/app/dashboard/DashboardPage.tsx
@@ -35,8 +35,7 @@ export function DashboardPage(props: { config: Config; basePath: string }) {
       <PageBody isScrollable>
         <Flex direction="column" gap="xxlarge">
           {user && <UserInfo user={user} manageAccount={!!cloudInfo} />}
-
-          <BranchSection config={props.config} />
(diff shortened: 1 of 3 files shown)


jossmac:
This will break core functionality. For example, if there's branch protection on `main` and a user is prompted to create a new branch they'll be stranded with no way to navigate.

branch protection dialog

Barruchou:
Hello @jossmac 

There is no config to change the default branch (other than "main")? In this case it would leave the developer the choice to have a "content" branch (for example) and continue to protect his "main" branch. Or allow you to define a fallback branch which would be used in the case of a protected "main".

What do you think ?

jossmac:
Hi @Barruchou, sorry I meant "default branch" but said `main` so it aligned with the screenshot. We'll look into an approach that allows consumers to configure a "content branch" or similar, but simply hiding UI that breaks functionality isn't a viable solution.

jossmac:
This is something we'd like to explore, if there's community interest. @Barruchou could you [create a discussion](https://github.com/Thinkmill/keystatic/discussions/categories/ideas)?

Barruchou:
And here it is: https://github.com/Thinkmill/keystatic/discussions/759

PR #771, opened 2023-11-21
Docs: diff indicators

jossmac:
Exclude diff indicators ("+" & "-") from selection; makes it easier to copy/paste.


simonswiss:
Works great!

This probably blows up the scope of this PR, but is there a way to *not* copy the lines starting with a `-`?

I tried copying the following diff from the docs:

```diff
export default defineConfig({
-  integrations: [react(), markdoc()],
+  integrations: [react(), markdoc(), keystatic()],
+  output: 'hybrid',
})
```

As expected, the result when I pasted was this:

```ts
export default defineConfig({
   integrations: [react(), markdoc()],
   integrations: [react(), markdoc(), keystatic()],
   output: 'hybrid',
})
```

I imagine most of the time when copying diff code, you'd only want to keep the `+` lines and not the `-` ones in your clipboard?

This is already a great improvement, so this comment should not prevent this PR from going through 👍

emmatown (approved):

jossmac:
> is there a way to not copy the lines starting with a -?

That should be possible, but could be unexpected behaviour for the user. I'll have a think about it.

jossmac:
> is there a way to not copy the lines starting with a -?

@simonswiss That should be possible, but would likely yield unexpected behaviour for the user. I'll have a think about it.

PR #774, opened 2023-11-22
Add Remix template + CLI Remix option

simonswiss:
- upgrades the `dev-projects/remix` to Remix v2 (necessary to satisfy `manypkg`'s version recommendations)
- adds a new `templates/remix` project
- adds a new `Remix` option to the Keystatic CLI that installs said template

CleanShot 2023-11-22 at 11 25 34@2x

I cannot test the CLI until the template is deployed/published, but it seems reasonably correctly setup to work.


templates/remix/package-lock.json
lukebennett88, line 1: Is this file supposed to be here?

simonswiss: hmmm, absolutely not — it's a side effect from copying the project over from a standalone thing. Good catch!

emmatown (approved):

PR #786, opened 2023-11-28
Field object layout

### LINKED ISSUE #669: Object field label + description gutter
The object field's label + description are too close to the enclosed fields.

Also consider increasing the gutter above and below the group, and potentially include some sort of delineating ornament indicating the relationship between enclosed fields like the native [fieldset](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset).

object field options example screen shot

jossmac:
Support "layout" prop on `fields.object()`. Also, add required indicators to relevant fields.

Resolves #669

### DIFF
--- a/.changeset/tasty-pillows-think.md
+++ b/.changeset/tasty-pillows-think.md
@@ -1,0 +1,5 @@
+---
+'@example/next-app': patch
+---
+
+Update "@example/next-app" config; introduce object layouts.
--- a/.changeset/empty-ligers-behave.md
+++ b/.changeset/empty-ligers-behave.md
@@ -1,0 +1,5 @@
+---
+'@keystatic/core': patch
+---
+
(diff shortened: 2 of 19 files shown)


.changeset/tasty-pillows-think.md
```diff
@@ -0,0 +1,5 @@
+---
+'@example/next-app': patch
```
emmatown, line 2: Just btw, the examples don't need changesets

emmatown (approved):

.changeset/tasty-pillows-think.md
jossmac: Worth removing?

emmatown: Not really, doesn't matter

PR #792, opened 2023-11-29
Editor prose styles

jossmac:
Inherit existing "prose" styles:

| Before | After |
|--------|--------|
| editor screenshot: before | editor screenshot: after |

Also added placeholder text "Start writing or press "/" for commands…" when the document is empty.


packages/keystatic/src/form/fields/markdoc/editor/schema.tsx
```diff
@@ -315,27 +279,6 @@ const nodeSpecs = {
     toDOM() {
       return blockquoteDOM;
     },
-    reactNodeView: {
```
jossmac, line 318: This is the only part i'm unsure about. Removing `reactNodeView` didn't seem to affect editor rendering but i'm definitely missing a bunch of context.

The wrapping `<div>` breaks margin expectations, though that's more of a shortcoming with the prose styles.

emmatown (approved):

PR #805, opened 2023-12-07
Support astro@4

emmatown:


packages/astro/src/index.ts
```diff
@@ -42,12 +42,16 @@ import "@keystatic/core/ui";
           vite,
         });
         injectRoute({
+          // @ts-ignore
           entryPoint: '@keystatic/astro/internal/keystatic-astro-page.astro',
```
simonswiss, line 46: Is leaving this here to support Astro 3 as well?

simonswiss (approved):

packages/astro/src/index.ts
emmatown: Yes

PR #836, opened 2023-12-30
Add link to the type signature page in Datetime API doc

armno:
Make the URL of the type signature in the [Datetime doc](https://keystatic.com/docs/fields/datetime) link to the actual page


emmatown (approved):
Thanks!

PR #842, opened 2024-01-08
Fix `fields.integer` allowing any number

### LINKED ISSUE #820: Field `integer` is not integer.
The field [integer](https://keystatic.com/docs/fields/integer) can contains decimal number. This is confusing IMHO.

Solutions:
1. Rename field name to `number` or `decimal`
2. Change behaviour of `integer` and add new field `number` or `decimal`

emmatown:
Fixes #820

### DIFF
--- a/design-system/pkg/src/number-field/NumberField.tsx
+++ b/design-system/pkg/src/number-field/NumberField.tsx
@@ -55,6 +55,7 @@ export const NumberField: ForwardRefExoticComponent<
       {...onlyStyleProps(props)}
       label={label}
       description={description}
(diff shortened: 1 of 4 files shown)


design-system/pkg/src/number-field/NumberField.tsx
```diff
@@ -55,6 +55,7 @@ export const NumberField: ForwardRefExoticComponent<
       {...onlyStyleProps(props)}
       label={label}
       description={description}
+      errorMessage={props.errorMessage}
```
jossmac, line 58: Oh god 😨 sorry I missed this…

PR #860, opened 2024-01-15
Use Node ESM in packages to allow using MDX packages

emmatown:


templates/nextjs/tsconfig.json
```diff
@@ -10,7 +10,7 @@
     "incremental": true,
     "esModuleInterop": true,
     "module": "esnext",
-    "moduleResolution": "node",
+    "moduleResolution": "bundler",
```
tyteen4a03, line 13: @emmatown this is a breaking change for people who have generated CJS projects before. Would it be possible to add a PSA somewhere?

PR #863, opened 2024-01-16
Remove react-transition-group

emmatown:
react-transition-group was causing module related problems and it was pretty simple to re-create the behaviour we were using with some state and an effect


design-system/pkg/src/overlays/Overlay.tsx
```diff
@@ -1,10 +1,17 @@
...
+  useState,
+} from 'react';
 import { Overlay as ReactAriaOverlay } from '@react-aria/overlays';
 import { OverlayProps } from '@react-types/overlays';
 
 import { KeystarProvider } from '@keystar/ui/core';
+import { cloneValidElement } from '../utils';
 
-import { OpenTransition } from './OpenTransition';
+const forceReflow = (node: HTMLElement) => node.scrollTop;
```
jossmac, line 14: Won't this just return a number? Not sure how this works, sorry.

If I understand the intent, it seems like idiomatic React would be to wrap the appropriate `setState()` call with [forceUpdate()](https://react.dev/reference/react/Component#forceupdate).

design-system/pkg/src/overlays/Overlay.tsx
```diff
@@ -14,36 +21,47 @@ export const Overlay = forwardRef(function Overlay(
...
+      if (isOpen) {
+        props.onEntered?.();
+      } else {
+        props.onExited?.();
+      }
+    }
+    if (isOpen === isOpenState) return;
+    hasCalledCompletedCallback.current = false;
+    if (isOpen) {
+      props.onEntering?.();
```
jossmac, line 47: Should this be `onEnter`?

design-system/pkg/src/overlays/Overlay.tsx
emmatown: Yeah, it triggers a side-effect, it's the same as what react-transition-group does: https://github.com/reactjs/react-transition-group/blob/2989b5b87b4b4d1001f21c8efa503049ffb4fe8d/src/utils/reflow.js

PR #886, opened 2024-01-18
feat(docs): add GitHub reader docs

tyteen4a03:


emmatown (commented):
This isn't really the right framing, the reader you use doesn't need to correspond to the storage option in your Keystatic config. For example, the Keystatic docs site uses cloud for storage and the file system reader and GitHub reader for different purposes

tyteen4a03:
> This isn't really the right framing, the reader you use doesn't need to correspond to the storage option in your Keystatic config. For example, the Keystatic docs site uses cloud for storage and the file system reader and GitHub reader for different purposes

Not sure I follow - the docs that predate my addition does not imply you need to use whatever's specified in your `keystaticConfig`.

tyteen4a03:
(re-requesting review because I noticed I forgot to update the frontmatter - oops)

emmatown:
I'm referring to this and etc:
> For `local` storage, import the `createReader` function, as well as your Keystatic config file:

It implies that `createReader` is just for `local` storage which it is not

tyteen4a03:
I think I see what you mean - I've updated the wording.

docs/src/content/pages/reader-api.mdoc
```diff
@@ -1,35 +1,62 @@
...
+1. The Keystatic config
+2. An options object containing:
+    - `repo`: The name of the content repository on GitHub (e.g. `Thinkmill/keystatic-data`)
+    - `token`: The Personal Access Token that allows read access to the repository. This is different from your GitHub App Client ID / Secret in `.env`.
+       For information on how to generate PATs, see [GitHub's documentation](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens).
+
+```javascript
+const reader = createGitHubReader(keystaticConfig, {
+    repo: 'Thinkmill/keystatic-data',
+    token: process.env.GITHUB_PAT, // or a hardcoded string
```
emmatown, line 56:
```suggestion
    token: process.env.GITHUB_PAT,
```

docs/src/content/pages/reader-api.mdoc
```diff
@@ -1,35 +1,62 @@
...
 The code can only run on the server, and not in the browser.
 {% /aside %}
 
 ## Usage
 
-First, import the `createReader` function, as well as your Keystatic config file:
+{% aside icon="⚠️" %}
+There is currently no support for creating readers for `cloud` storage.
+{% /aside %}
+
```
emmatown, line 18:
```suggestion
```

emmatown (approved):

docs/src/content/pages/reader-api.mdoc
tyteen4a03: I don't mind leaving this out but I feel people who use `cloud` will ask questions either way.

docs/src/content/pages/reader-api.mdoc
tyteen4a03: No objections leaving this out.

docs/src/content/pages/reader-api.mdoc
emmatown: Creating a reader with a cloud config works totally fine?

tyteen4a03: I don't see a `createCloudReader` that's why I thought it doesn't work. Not that I've looked into how `cloud` works though.

PR #910, opened 2024-01-23
Reduce icon bundle size

emmatown:


jossmac:
Probably won't be an issue but the API supported passing any SVG with a 24x24 viewbox for flexibility. Not sure if the "src" prop really makes sense now.

PR #913, opened 2024-01-25
Add `parseSlugForSort` option to collections

emmatown:


JedWatson (approved):

tordans:
Can someone tell me a bit more about the use case for this option? The docs are a bit self referencing …
I especially wonder, why only the slug is passed as function param and not other fields of the collection.
Maybe someone can share a few use cases and I can add a PR to extend the docs on this.

PR #934, opened 2024-02-06
fix(docs): invalid JSON doesn't match real type

pnodet:
Actual `DataFormat` is 'json' | 'yaml'


emmatown (approved):

simonswiss:
Thanks for this @pnodet 🤙

PR #966, opened 2024-02-24

pkellner:
went to far down with the ../../ stuff. Just one too many


simonswiss (approved):

simonswiss:
I think this is actually correct? The `route.ts` file is nested 5 directories deep inside the `src` dir, so we need to go back 5x to be in the project root where the Keystatic config is.

PR #967, opened 2024-02-24

pkellner:
With app router in next, the index page is /keystatic/page.ts, not keystatic/keystatic.ts


simonswiss:
This is not used as a page here, this is a component that is then added to the `layout`, so the current implementation is correct.

PR #970, opened 2024-02-25

pkellner:
Content needs to store images in /public folder or they won't get displayed.


simonswiss (approved):
Good point, I have meant to update this in a while.

PR #973, opened 2024-02-27
feat(astro): make integration manage custom host, cleanup

florian-lefebvre:
Now, updating the dev script in astro projects is no longer required! It will be managed by the integration. I also simplified the integration code

### DIFF
--- a/packages/astro/internal/keystatic-api.js
+++ b/packages/astro/internal/keystatic-api.js
@@ -1,5 +1,8 @@
 import { makeHandler } from '../dist/keystatic-astro-api.js';
+// eslint-disable-next-line import/no-unresolved
 import config from 'virtual:keystatic-config';
 
 export const all = makeHandler({ config });
 export const ALL = all;
+
+export const prerender = false;
--- a/packages/astro/internal/keystatic-api.js
+++ b/packages/astro/internal/keystatic-api.js
@@ -1,8 +1,5 @@
 import { makeHandler } from '../dist/keystatic-astro-api.js';
-// eslint-disable-next-line import/no-unresolved
 import config from 'virtual:keystatic-config';
 
 export const all = makeHandler({ config });
 export const ALL = all;
-
-export const prerender = false;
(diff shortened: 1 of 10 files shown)


simonswiss:
This looks 🔥

Love that you updated the docs as well — thank you!

I'll leave it to @emmatown to review, but the prospect of not having to add `--host 127.0.0.1` on every project is awesome!

packages/astro/internal/keystatic-api.js
```diff
@@ -1,8 +1,5 @@
 import { makeHandler } from '../dist/keystatic-astro-api.js';
-// eslint-disable-next-line import/no-unresolved
 import config from 'virtual:keystatic-config';
 
 export const all = makeHandler({ config });
 export const ALL = all;
-
-export const prerender = false;
```
florian-lefebvre, line 8: Since `prerender` is specified in `injectRoute`, this is redundant

packages/astro/internal/keystatic-api.js
```diff
@@ -1,8 +1,5 @@
 import { makeHandler } from '../dist/keystatic-astro-api.js';
-// eslint-disable-next-line import/no-unresolved
```
florian-lefebvre, line 2: Virtual import is now typed (in the monorepo only) using `virtual.d.ts`

packages/astro/src/index.ts
```diff
@@ -1,46 +1,49 @@
...
+import { mkdirSync, writeFileSync } from 'node:fs';
 
 export default function keystatic(): AstroIntegration {
   return {
     name: 'keystatic',
     hooks: {
-      'astro:config:setup': async ({ injectRoute, updateConfig, config }) => {
-        if (config.output !== 'hybrid') {
+      'astro:config:setup': ({ injectRoute, updateConfig, config }) => {
+        if (config.output === 'static') {
```
florian-lefebvre, line 9: As long as `prerender` can be `false`, it should be usable by the users

simonswiss: Yup, I think that makes sense to allow both `hybrid` and` server` — as long as the Keystatic routes themselves are not static 🤙

emmatown (approved):

PR #980, opened 2024-03-02
fix: missing brace in installation-next-js.mdoc

pnodet:


simonswiss (approved):
Thanks for this!

pnodet:
@simonswiss Sure!

PR #1051, opened 2024-04-03
feat: allow customizing commit messages

### LINKED ISSUE #1033: allow customizing commit messages
when using `github` storage mode, it would be great to be able to customize the commit messages.

i think these are currently set [here](https://github.com/Thinkmill/keystatic/blob/main/packages/keystatic/src/app/updating.tsx#L216) and [here](https://github.com/Thinkmill/keystatic/blob/main/packages/keystatic/src/app/updating.tsx#L404).

stefanprobst:
closes #1033

this allows providing custom commit messages via the `storage` config in `keystatic.config.ts`:

```ts
{
  kind: "github",
  repo: { owner: "stefanprobst", name: "keystatic-test" },
  commitMessages: {
    update({ path }) {
      return `content: update ${path}`
    },
    delete({ path }) {
      return `content: delete ${path}`
    },
  }
}
```


emmatown:
Thanks for the PR! Unfortunately we can't accept it because we're planning to change how committing works so many changes can be batched into a single commit so these messages wouldn't make sense. Those changes will allow users to customise the commit message in the UI though and we could look into customising the defaults once that is in.

PR #1053, opened 2024-04-04
Change keystatic.config.ts in docs example to keystatic.config.tsx where JSX is used

### LINKED ISSUE #1052: Brand UI Mark Example don't works.
In that doc page there is an example how to change the logo alias `mark` of the keystatic cms.

Here is the link: https://keystatic.com/docs/user-interface

But the example doesn't works.. 

P.S. I work with an a Astro project if it can change something.

Here the example: (I tried in few way with different image)

```javascript
// keystatic.config.ts
import { config } from '@keystatic/core'

export default config({
  ...
  ui: {
    brand: {
      name: 'Your brand',
      mark: ({ colorScheme }) => {
        let path = colorScheme === 'dark'
          ? '//your-brand.com/path/to/dark-logo.png'
          : '//your-brand.com/path/to/light-logo.png';
        
        return <img src={path} height={24} />
      },
    },
  },
})
```

---

Here is an example that works:

```javascript
// keystatic.config.ts
import { config } from '@keystatic/core'
import LogoDark from './public/logo-white.svg'
import LogoWhite from './public/logo-dark.svg'
import React from 'react';

export default config({
  ...
  ui: {
    brand: {
      name: 'Your brand',
      mark: ({ colorScheme }) => {
        let themeLogo = colorScheme === 'dark' ? LogoDark.src : LogoWhite.src
        
        return React.createElement("img", {src: themeLogo, width: 30})
      },
    },
  },
})
```

---

Maybe there is a bug to fix for the actual example to works?
Well there is my workaround if it is.

emmatown:
Fixes #1052

### DIFF
--- a/docs/src/content/pages/user-interface.mdoc
+++ b/docs/src/content/pages/user-interface.mdoc
@@ -1,7 +1,8 @@
 ---
 title: User interface
 summary: >-
-  Configure parts of the Admin UI to improve the experience of your content editors.
+  Configure parts of the Admin UI to improve the experience of your content
+  editors.
 ---
 Configure parts of the Admin UI to improve the experience of your content editors. Making the interface familiar to your editors will help them get started quickly and feel at home.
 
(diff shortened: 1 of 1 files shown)


t1gu1:
Not sure of that commit.
Read that comment: https://github.com/Thinkmill/keystatic/issues/1052#issuecomment-2048462647

PR #1061, opened 2024-04-09
Array field required indicator

### LINKED ISSUE #982: display required asterisk for array fields with min-length > 0
hi, i have the following array field in a schema, and think it would be nice to also display a "required" indicator (asterisk), when the array field has a field validation of `{ length: { min: 1 } }`:

```ts
tags: fields.array(
  fields.text({
    label: "Tag",
    validation: { isRequired: true },
  }),
  {
    label: "Tags",
    validation: { length: { min: 1 } },
  }
),
```

jossmac:
Resolves #982

UX and A11Y improvements for `fields.array()` and `fields.blocks()` interfaces:
- Include the "required" indicator when min length validation provided by consumer.
- Connect the validation message to the group accessibly.

Unrelated:
- Minor improvement for `fields.url()` mobile soft-keyboard experience.

### DIFF
--- a/packages/keystatic/src/form/fields/array/ui.tsx
+++ b/packages/keystatic/src/form/fields/array/ui.tsx
@@ -4,65 +4,74 @@ import { ItemDropTarget, useDragAndDrop } from '@keystar/ui/drag-and-drop';
 import { FieldLabel, FieldMessage } from '@keystar/ui/field';
 import { Icon } from '@keystar/ui/icon';
 import { trash2Icon } from '@keystar/ui/icon/icons/trash2Icon';
-import { Flex } from '@keystar/ui/layout';
+import { VStack } from '@keystar/ui/layout';
 import { Item, ListView } from '@keystar/ui/list-view';
 import { Content } from '@keystar/ui/slots';
 import { Tooltip, TooltipTrigger } from '@keystar/ui/tooltip';
 import { Heading, Text } from '@keystar/ui/typography';
 import { useLocalizedStringFormatter } from '@react-aria/i18n';
-
-import { previewPropsToValue, valueToUpdater } from '../../get-value';
-import { createGetPreviewProps } from '../../preview-props';
+import { useField } from '@react-aria/label';
+import { useId, Key, useMemo, useState } from 'react';
 
 import l10nMessages from '../../../app/l10n/index.json';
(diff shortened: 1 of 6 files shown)


packages/keystatic/src/form/fields/array/ui.tsx
```diff
@@ -4,78 +4,89 @@ import { ItemDropTarget, useDragAndDrop } from '@keystar/ui/drag-and-drop';
...
+          setModalState({ state: 'new' });
         }}
         alignSelf="start"
       >
         {stringFormatter.format('add')}
       </ActionButton>
       <ArrayFieldListView
         {...props}
-        labelId={labelId}
+        aria-label={props.schema.label}
```
jossmac, line 82: Switch to `aria-label` so the "required" text isn't included.

packages/keystatic/src/form/fields/array/ui.tsx
```diff
@@ -4,78 +4,89 @@ import { ItemDropTarget, useDragAndDrop } from '@keystar/ui/drag-and-drop';
...
+  const minLength = props.schema.validation?.length?.min ?? 0;
+  const formId = useId();
   const stringFormatter = useLocalizedStringFormatter(l10nMessages);
+  const errorMessage = useArrayFieldValidationMessage(props);
+  const {
+    descriptionProps,
+    errorMessageProps,
+    fieldProps: groupProps,
+    labelProps,
+  } = useField({
```
jossmac, line 44: Prefer `useField()` here; let it handle the `aria-describedby` logic when combination of some, none or all description + validation message present.

packages/keystatic/src/form/fields/array/ui.tsx
```diff
@@ -409,9 +397,7 @@ export function ArrayFieldListView<Element extends ComponentSchema>(
               >
                 <Icon src={trash2Icon} />
               </ActionButton>
-              <Tooltip tone="critical">
-                {stringFormatter.format('delete')}
-              </Tooltip>
+              <Tooltip>{stringFormatter.format('delete')}</Tooltip>
```
jossmac, line 400: Unnecessary critical tone; it's not a destructive action. Prefer "remove" but we don't have the l10n strings.

emmatown (approved):

PR #1064, opened 2024-04-10
Fix loading spinner replacing the page for a short amount of time when saving an entry

emmatown:


packages/keystatic/src/app/ItemPage.tsx
```diff
@@ -882,37 +882,53 @@ function ItemPageWrapper(props: ItemPageWrapperProps) {
...
+          doc = new Y.Doc();
+          yjsInfo.data.set(key, doc);
+        }
+        const data = doc.getMap('data');
+
+        return data;
+      })();
+    }, [isItemDataLoading, isItemNotFound, key, yjsInfo])
   );
+  useMemo(() => {
```
jossmac, line 910: Interesting, I've never seen `useMemo` w/o a returned value. Is there a benefit over `useEffect` here?

jossmac (approved):

packages/keystatic/src/app/ItemPage.tsx
emmatown: I want it to happen in render rather than in an effect

PR #1083, opened 2024-04-21
Add basic support for custom CSS in the app

filipesmedeiros:
Hey! This adds basic support for custom CSS in the app, to allow some quirky behaviour (that I can exemplify if needed).

For example, this allows adding something like a utility-based CSS library (by precompiling the CSS, of course).

Of course this can be improved in the future, but for now it already covers a lot of cases even with the simplest implementation of all!

The next steps would be probably to allow to introduce things like TailwindCSS in the build step of the Admin UI specifically, or maybe something else I can't think of right now, I'm tired ahah

Hope you enjoy this simple solution :) Let me know if you want me to change anything.

### DIFF
--- a/.changeset/selfish-items-double.md
+++ b/.changeset/selfish-items-double.md
@@ -1,0 +1,5 @@
+---
+'@keystatic/core': minor
+---
+
+Add option to include custom CSS in admin UI
--- a/packages/keystatic/src/app/provider.tsx
+++ b/packages/keystatic/src/app/provider.tsx
@@ -205,6 +205,7 @@ export default function Provider({
           href="https://fonts.googleapis.com/css2?family=Inter:wght@400;500;600;700&display=swap"
           rel="stylesheet"
         />
+        <style>{config.customCSS}</style>
         <UrqlProvider value={useMemo(() => createUrqlClient(config), [config])}>
           {children}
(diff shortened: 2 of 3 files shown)


filipesmedeiros:
I now realise I forgot to add tests... 😓

emmatown:
Thanks for the PR but this doesn't really make sense to merge, since Keystatic is rendered inside your own project, you can add a `<style>` tag or insert CSS any other way without involving Keystatic itself

filipesmedeiros:
@emmatown oh! Now I feel dumb ahah do you have maybe a pointer on how to do this? I don't really see how to access the Keystatic pages to add a layout, for example, but I might be missing something. Thanks!

emmatown:
I'm guessing you're using Astro since in Next and Remix, the pages are just there so for Astro, you could use https://docs.astro.build/en/reference/integrations-reference/#injectscript-option to insert a CSS import in every page (since you mentioned wanting to Tailwind in your PR description, I believe you can just use the Astro Tailwind integration which uses `injectScript` and you can just use Tailwind everywhere). You could also have a look at what `@keystatic/astro` is doing under the hood and create those pages manually and customise them however you want.

filipesmedeiros:
@emmatown Thank you so much :)

PR #1097, opened 2024-04-25
Store search query state in URL

jplhomer:
This PR stores the search query in the URL as a `search` query string, and it pre-populates the initial value of the collection page with the current search query.

First contribution here, so please forgive any formatting or code pattern violations 😄

PR #1105, opened 2024-04-26
Track sort descriptor in url

jplhomer:
Coming hot off #1097, this PR stores the column sort descriptor in a URL param 😄


jplhomer:
I'm starting to think about some alternatives to this...

The real problem I'm trying to solve is that I'd love to apply a **default sort** to my collection of posts, e.g. reverse chronological by `publishDate`. Having sort params persisted in the URL is nice, but still doesn't totally solve that problem.

Perhaps adding a `sort` callable function option or `defaultSortBy + defaultSortDir` options to the collection definition would be a better approach?

jplhomer:
OK, I PRed this alternate approach here, and I think I like it better than simply tracking params in the URL https://github.com/Thinkmill/keystatic/pull/1110

PR #1110, opened 2024-04-29
New `collection` option: `defaultSort`

jplhomer:
This PR is an alternative to #1105 which provides an option for the default sort to be used by a Collection list view.

I chose the following API:

```ts
defaultSort: SortDescriptor
```

This provides a way for users to default to e.g. a reverse chronological order on their collections list view for a set of blog posts.


packages/keystatic/src/config.tsx
```diff
@@ -187,6 +189,9 @@ export function collection<
         ? K & string
         : never;
     }[keyof Schema][];
+    defaultSort?: {
+      column?: keyof Schema;
```
jplhomer, line 193: Caveat: this allows _any_ field from the schema to be used as the sort column, but I couldn't figure out how to tell TypeScript that I only want to allow the values passed in the sibling `columns` property 😄 

Kind of a YOLO doesn't matter thing probably. But we could also rename it to `field` if we wanted to be pedantic.

jossmac:
FYI: Something similar is planned as part of a larger "columns definition" body of work, which is planned as a [roadmap item](https://github.com/orgs/Thinkmill/projects/6?pane=issue&itemId=59412408) to be picked up in the near future.

jplhomer:
Closing this to defer to this [roadmap item](https://github.com/orgs/Thinkmill/projects/6?pane=issue&itemId=59412408)

jasikpark:
Is there a way to track that roadmap item? I'm interested in seeing column sort happening! 💜

PR #1138, opened 2024-05-14
Docs: Added Next.js recipe for disabling admin in production

funabab:
Hello everyone 👋

I was integrating keystatic into a Next.js project and noticed there was only a recipe for disabling admin UI in production for Astro.

I updated the docs to show how to disable (404) `keystatic` routes in production with Next.js. Hoping someone will find it helpful 🙏.


simonswiss:
Nice one, thanks for that! This has been requested a few times, so it indeed comes handy 🙌

Issue #401, opened 2023-07-03
Improve Locale support


JedWatson:
We've added initial support for interface localisation but it's incomplete / broken for non-English users, specifically:

* Not all strings are localised so the UI is a mix of English and the Browser's locale
* The translated strings are probably wrong, they're currently sourced from Google Translate
* You can't override the Locale

To make this _less broken_ we should do a few things:

* Default the Locale specifically to English and make translations opt-in so people can test them (also allow them to be set in config)
* Make all the strings locale-based, removing the hard-coded English parts
* Document the current state and ask for help translating strings

Later when we have more confidence in our translation support, we could (re)introduce the option to use the browser's locale so that implementations of Keystatic with users who have different locales see their own, but right now that would lead to a broken experience on a per-end-user basis so let's leave it with the config to set globally for each project.

PR #1171, opened 2024-06-06
fix(i10n): update uk-UA translations

orlovol:
Hi! 👋
Just starting with keystatic+Astro, and noticed that translations could be improved.
Noticed there are plans to improve locales in #401 — can note down untranslated strings as I encounter them & come up with uk-UA translation for them.


emmatown (approved):
Thanks!

PR #1246, opened 2024-07-25
fields/slug: Make `pattern` optional for fields slug

tordans:
The `pattern` was introduce in https://github.com/Thinkmill/keystatic/pull/1228.

However it is a required field for `slug`s. See also https://github.com/Thinkmill/keystatic/commit/36ea216c4a473eeab60aff41cb9dfabfb45127b8#r144629665

The way I read the code is, that the code actually expect the pattern to be optional by using `?.` and setting a default of `undefined`. The pattern on the `text` field is optional.

I am assuming the optionality was missing for the slug field. This PR adds it.

### DIFF
--- a/.changeset/healthy-pets-act.md
+++ b/.changeset/healthy-pets-act.md
@@ -1,0 +1,5 @@
+---
+"@keystatic/core": patch
+---
+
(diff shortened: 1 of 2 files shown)


emmatown (approved):
Thanks!

PR #1249, opened 2024-07-30
Collection view: table columns definition

jossmac:
Implements the [columns definition](https://github.com/orgs/Thinkmill/projects/6/views/1?pane=issue&itemId=59412408) proposal, while maintaining the original API.

Thank you [jplhomer](https://github.com/jplhomer) for your work in #1110, we appreciate your patience on this feature.

### DIFF
--- a/docs/src/content/pages/collections.mdoc
+++ b/docs/src/content/pages/collections.mdoc
@@ -1,16 +1,16 @@
 ---
 title: Collections
 summary: >-
-  Think of a collection as anything you'd want multiple instances of. A series
+  Think of a collection as anything you’d want multiple instances of. A series
   of blog posts, cooking recipes, or testimonials from happy customers.
 ---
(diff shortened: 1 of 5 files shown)


docs/src/content/pages/collections.mdoc
```diff
@@ -37,29 +37,84 @@ export default config({
 
 ### Columns
 
-`columns` — show additional fields in the collection list view. 
+By default, only the “slug” of each entry is displayed in the collection table. You can show additional fields by passing a `columns` option.
```
emmatown, line 40: > By default, only the “slug” of each entry is displayed in the collection table. You can show additional fields by passing a `columns` option.

This is no longer true with this PR, `columns` now removes the slug instead of just adding other columns. I don't think we should land this without preserving the ability to show the slug and other fields.

emmatown (changes requested):

PR #1275, opened 2024-08-26
DS fixes and updates

jossmac:
Some things I noticed while integrating with Keystone.

Updates:

- Support "isPending" prop on `Button`
- Support "low" prominence `Checkbox`
- Emphasise "selected" state on `ActionButton`
- More prominent `ActionBar`
- Increase `TextArea` min-height to 3 lines

Fixes:

- Allow "focus" method on `Picker` ref
- Defensive "current" selector on `NavItem` styles
- Fix text truncation on `Picker` selected text
- Clear slots of `Content` children—resolves issue with `Calendar` elements within `Dialog` receiving incorrect props
- Fix issue with `Tray` when "size" provided to `Dialog` component


design-system/pkg/src/picker/Picker.tsx
```diff
@@ -57,6 +58,13 @@ function Picker<T extends object>(
   let popoverRef = useRef<HTMLDivElement>(null);
   let triggerRef = useRef<HTMLButtonElement>(null);
   let listboxRef = useRef<HTMLDivElement>(null);
+  // @ts-expect-error FIXME: not sure how to properly resolve this type issue
+  useImperativeHandle(forwardedRef, () => ({
+    ...triggerRef.current,
+    focus() {
+      state.setFocused(true);
+    },
+  }));
```
emmatown, line 67: This doesn't seem right, `forwardedRef` is already being passed to `FieldPrimitive`? And what is being focused here? I assume not the trigger button since then overriding it here wouldn't change anything right? Also, not sure about spreading an `HTMLElement` like this since i think lots of methods wouldn't work since the `this` wouldn't be right + it would unnecessarily evaluate a bunch of getters, maybe something like this might make sense?
```ts
useImperativeHandle(forwardedRef, () => ({
  get trigger() {
    return triggerRef.currrent;
  },
  focus() {
    state.setFocused(true);
  },
}));
```

jossmac: I'll remove this from the PR and we can revisit later 👍

jossmac: A bit of context: Internally, spectrum [packs and unpacks refs](packages/@react-spectrum/utils/src/useDOMRef.ts) with `useDOMRef`, which lets them target the focusable element. I'd like to avoid that awkwardness but may have been a bit hasty in my implementation.

PR #1308, opened 2024-09-13
Fix error when rendering empty cell in `<TableView>`

emmatown:


jossmac:
Good catch, thank you! Sorry about that

jossmac (commented):
Good catch, thank you! Sorry about that, was definitely rushing a bit.

PR #1345, opened 2024-11-03
Misc. component lib fixes and improvements

jossmac:
Address quirks and bugs encountered while implementing Keystone admin UI.

### DIFF
--- a/design-system/pkg/src/calendar/CalendarCell.tsx
+++ b/design-system/pkg/src/calendar/CalendarCell.tsx
@@ -122,6 +122,7 @@ export function useCellStyles(props: Partial<CellStyleProps> = {}) {
       trimBooleanKeys: true,
     }),
     className: css({
+      boxSizing: 'content-box',
       height: cellSize,
       padding: cellPadding,
       position: 'relative',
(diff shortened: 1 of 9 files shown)


design-system/pkg/src/calendar/CalendarCell.tsx
```diff
@@ -122,6 +122,7 @@ export function useCellStyles(props: Partial<CellStyleProps> = {}) {
       trimBooleanKeys: true,
     }),
     className: css({
+      boxSizing: 'content-box',
```
jossmac, line 125: Explicit declaration overrides global reset (if any).

design-system/pkg/src/number-field/NumberField.tsx
```diff
@@ -50,23 +58,25 @@ export const NumberField: ForwardRefExoticComponent<
 
   return (
     <TextFieldPrimitive
+      ref={inputRef}
```
jossmac, line 61: Ref wasn't passed on properly + add missing props `contextualHelp`, and `isRequired`.

emmatown (approved):
Needs a changeset but otherwise looks fine

PR #1356, opened 2024-11-20
Docs: github-mode: Hint at different env var names

tordans:
The `NEXT_PUBLIC_KEYSTATIC_GITHUB_APP_SLUG` is only the right name for Next.JS integrations.
This PR adds a hint for Astro users to look out for a different variable.

---

Side note: I wonder if this condition should be extended to handle both variables as well https://github.com/Thinkmill/keystatic/commit/618ba4aa2932bd9b1c6e73d72ff24bf1e550b4ff#diff-8b4b55d3c75b76109c771dcd8a01e1b1339c07b753bb409b93e71df2f9d4d2cfR91


docs/src/content/pages/github-mode.mdoc
```diff
@@ -131,4 +131,4 @@ This segment of the [Keystatic Mini-Course on YouTube](https://www.youtube.com/p
 
 {% embed
    mediaType="video"
-   embedCode="<iframe src=\"https://www.youtube.com/embed/BAnfePGzkbg?si=mDDK52Nlhg4v1v9x\" title=\"YouTube video player\" frameborder=\"0\" allow=\"accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share\" allowfullscreen></iframe>" /%}
\ No newline at end of file
+   embedCode="<iframe src=\"https://www.youtube.com/embed/BAnfePGzkbg?si=mDDK52Nlhg4v1v9x\" title=\"YouTube video player\" frameborder=\"0\" allow=\"accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share\" allowfullscreen></iframe>" /%}
```
tordans, line 134: FYI: This change happened automatically when editing via Github UI

emmatown:
> Side note: I wonder if this condition should be extended to handle both variables as well https://github.com/Thinkmill/keystatic/commit/618ba4aa2932bd9b1c6e73d72ff24bf1e550b4ff#diff-8b4b55d3c75b76109c771dcd8a01e1b1339c07b753bb409b93e71df2f9d4d2cfR91

No, this was a specific thing to give users a hint at what might be going wrong because that commit required updates to both `@keystatic/core` and `@keystatic/next` but not other framework packages

emmatown (approved):

PR #1383, opened 2024-12-22
Fix polish translations

Coolxer:
I fixed errors in Polish translations, including changing ‘Dodać’ to ‘Dodaj’, and corrected several other inaccuracies.


emmatown (approved):
Thanks!

PR #1401, opened 2025-02-09
Update `datetime` field to correctly store in UTC

### LINKED ISSUE #1400: `datetime` field: incorrect UTC converstoin
The `datetime` field currently captures dates in local time, using an `datetime-local` <input> type. However, these values are as UTC by appending a "Z", without correctly adjusting the local time to UTC:

https://github.com/Thinkmill/keystatic/blob/4c4b0efa8ecfc45053dee992f7ceb8566c520ede/packages/keystatic/src/form/fields/datetime/index.tsx#L64

This implementation requires content users to manually adjust `Date` objects created from these strings to the time zone of the content editor, which may not always be known. 

Instead, the `datetime` fields should convert its data from local time to UTC before serialization, to ensure predictable alignment  between content editors and content users.

### Overview:
- The datetime-local input type captures dates in local time.
- The current implementation appends a "Z" to the local time, indicating UTC, without converting the local time to UTC.
- This can lead to inconsistencies and unexpected behavior when the serialized date is interpreted as UTC.
- The datetime component should convert the local time to UTC before appending the "Z" and serializing the date.
- This change will ensure that the serialized date is consistently interpreted as UTC, leading to more predictable behavior.

mackeyguenther:
fixes #1400 

1. update `datetime` field (UI component & validation) to serialize/deserialize the value in UTC time, and render the value in the user's local timezone
2. update `datetime` documentation to make it clear that `datetime` stores its value in UTC time, and displays it in the user's local timezone in the editing interface

### DIFF
--- a/docs/src/content/pages/fields/datetime.mdoc
+++ b/docs/src/content/pages/fields/datetime.mdoc
@@ -4,7 +4,12 @@ summary: The datetime field stores a Datetime string.
 ---
 {% field-demo field="datetime" /%}
 
-The `datetime` field stores a Datetime string, collected from an `<input type="datetime-local" />` form field.
+The `datetime` field stores an [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) formatted datetime string in UTC time (i.e. YYYY-MM-DDTHH:MM:SS.000Z), collected from an `<input type="datetime-local" />` form field. 
+
+{% aside icon="☝️" %}
+Datetimes are stored in UTC time, but displayed/edited in the user's local timezone.
+{% /aside %}
+
 
 ## Usage example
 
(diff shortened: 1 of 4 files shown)


kylejrp:
@BarnabyBishop @emmatown @JedWatson Happy holidays, hope you all are doing well. 🎄 Would love to see this reviewed and get merged into the next release!

This is causing headaches for getting the local timezone right. If I write the date `2025-12-28 22:20` (10:20PM) in the Keystatic UI while in pacific time, it gets written in the frontmatter as `2025-12-28T22:20:00.000Z` which then incorrectly gets rendered by my site as `December 28th 2025, 2:20:00pm` which is off by an entire 8 hours. 😞

Issue #1410, opened 2025-02-20
Support for `keystatic.config.tsx` in Astro


afonsojramos:
It seems that the Astro integration only looks for `keystatic.config.ts`. It is easily fixable by exporting from a `keystatic.config.tsx` and reexporting from the `keystatic.config.ts` but it is a unnecessary workaround.

afonsojramos:
Having iterated over this, I ended up with a better alternative which is to just to export the component somewhere else and just import it in the `.ts` config.

```ts
mark: ({colorScheme}) => logo({colorScheme}),
```

emmatown:
I'm unable to reproduce this, example of using `keystatic.config.tsx` with the Astro integration: https://stackblitz.com/edit/withastro-astro-nenbjwqd?file=keystatic.config.tsx

afonsojramos:
I'm not being able to either anymore... Not sure what was happening here 😓

PR #1411, opened 2025-02-23
docs(mark): add astro exception for ui config

afonsojramos:
While #1410 is not solved, this should help anyone facing this issue.


afonsojramos:
Closing as #1410 wasn't being able to be reproduced.

PR #1415, opened 2025-03-02
Fix ActionGroup always collapsed button label

gautamsi:
I was working with extending the Keystone custom pages for List view with custom actions.
Seems like `ActionGroup` always collapsed and does not allow me adding visible buttons in the `ActionBar`.

this fixes the behavior which allows me use `Actionbar` with multiple items in the `ActionGroup` shown with the same key `buttonLabelBehavior`

default value is `collapse` for `ActionBar` anyways

before:
image

after:
image

### DIFF
--- a/.changeset/good-files-teach.md
+++ b/.changeset/good-files-teach.md
@@ -1,0 +1,5 @@
+---
+'@keystar/ui': patch
+---
+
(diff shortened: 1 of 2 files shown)


jossmac (approved):
Good call 🚀

PR #1485, opened 2025-12-27
feat(editor): add support for mailto: and tel: links

williamwmarx:
Expand URL pattern in document and markdoc editors to accept mailto: and tel: protocols in addition to http:// and https://. This allows users to paste email addresses and phone numbers as hyperlinks.


Greenheart (approved):
Thanks - this is a good improvement! We've noticed this problem in our projects too.

emmatown (approved):

PR #1525, opened 2026-03-31
Feature: Display image thumbnails in collection list views

mrprkr:
## Summary
- Adds thumbnail rendering for image fields when included in a collection's `columns` configuration
- Asset fields with image file extensions are detected and displayed as 32x32 rounded thumbnails instead of raw file paths
- Falls back gracefully to showing the path string if the image fails to load

## Changes
- **`packages/keystatic/src/app/CollectionPage.tsx`** — Added `ImageThumbnail` component, `isImageColumn()` and `isImagePath()` helpers, updated cell rendering logic
- **`dev-projects/next-app/keystatic.config.tsx`** — Added `columns` config with `heroImage` to posts collection for testing; configured `publicPath` and `directory` on the image field

## Test plan
- [ ] Create a collection with an `image` field and include it in the `columns` array
- [ ] Add entries with images and verify thumbnails appear in the collection list
- [ ] Verify entries without images show an empty cell
- [ ] Verify non-image asset fields (e.g. `file`) still display as text
- [ ] Verify broken image paths fall back to showing the path string


packages/keystatic/src/form/api.tsx
```diff
@@ -48,6 +48,7 @@ export type BasicFormField<
     parse(value: FormFieldStoredValue): ReaderValue;
   };
   label?: string;
+  Cell?: ComponentType<{ value: any }>;
```
emmatown, line 51: `value` should be `ParsedValue` here and etc.

